PR #4, opened 2023-08-25
fix(publish-techdocs): remove cache from setup-python

zzehring:
Setting this pip cache requires a `requirements.txt` file somewhere in the project or the step fails. Remove the `cache` input as not all projects will have a `requirements.txt`.


zzehring:
Example failure here: https://github.com/grafana/grafana-backstage/actions/runs/5976803372/job/16215439067

PR #5, opened 2023-08-25
fix(publish-techdocs): remove `cache` from `setup python` step name

zzehring:
Remove `cache` from the step name as we are no longer setting up cache.

### DIFF
--- a/.github/workflows/publish-techdocs.yaml
+++ b/.github/workflows/publish-techdocs.yaml
@@ -38,7 +38,7 @@ jobs:
       - name: Install techdocs-cli
         run: sudo npm install -g @techdocs/cli
 
-      - name: setup python cache
+      - name: setup python
         uses: actions/setup-python@v2
         with:
           python-version: 3.11


iainlane (approved):
why aren't we?

zzehring:
See https://github.com/grafana/shared-workflows/commit/040c7ef79b820444cca5bd940663fefef753b651

iainlane:
Can we get required reviews for this repo?

iainlane:
I see, alright, no worries. I just wanted to explore if we can keep it in some way (e.g. by downloading the techdocs-cli package in one step, setting up the cache, then installing it - it's really about caching its dependencies and not the project's own ones). But we can try that on the side too.

zzehring:
Yea agreed. We could probably set up a dummy `requirements.txt` for pip cache in the workflow and cleanup afterwards. That would ensure we have the file and get cacheing. 

I'd also want to look at using an image for this stuff and custom action. There was some work with that here: https://github.com/grafana/grafana-backstage/blob/main/actions/publish-techdocs/action.yml

PR #6, opened 2023-09-01
Bump python action

tolzhabayev:
SCR-20230901-m2j

This add a dependabot workflow that should lead to opening a PR to bump the python workflow that will eventually make this warning go away once merged.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "monthly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,4 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every week
       interval: "weekly"
+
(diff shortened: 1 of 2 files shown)


tolzhabayev:
> WDYT about adding Dependabot instead, then we can use this bump as a test to see if it works? Would you be willing to do that?

Alright, updated https://github.com/grafana/shared-workflows/pull/6/commits/ea55a172408f5895ab57320c1a228d258c43502a

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "monthly"
```
zzehring, line 10: Comment says `every week` but interval is `monthly`, so need to update one of those. I'm thinking `weekly` would be better but open to opinions.

zzehring, line 10: We should add a newline at EOF.
```suggestion
      # Check for updates to GitHub Actions every week
      interval: "monthly"
      
```

tolzhabayev: You are absolutely, right, I forgot to change the comment. But now changed it to weekly so it fits.

tolzhabayev: Added, thanks!

zzehring (approved):
LGTM! :shipit:

PR #9, opened 2023-09-13
Use workload identity federation instead of static SA credentials

iainlane:
We've just enabled support for this. Use it in downstream consumers -
gets rid of static service account credentials.


julienduchesne (approved):

zzehring (approved):
:shipit:

PR #15, opened 2023-11-22
Add support for triggering Argo workflows

iainlane:
Here we add a small tool wrapping the `argo` CLI, a Docker image to build and push it, and an action which uses it.

These let us trigger Argo workflows on our dev and prod instances. Credentials to trigger are scoped to only be requestable by the listed repository, and they can only trigger workflows in their own namespace too. Secrets are not available to forks.


zzehring (commented):
Very cool! Definitely gonna be useful. Comments in line.

actions/trigger-argo-workflow/action.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+  steps:
+    - name: Checkout
+      env:
+        action_repo: ${{ github.action_repository }}
+        action_ref: ${{ github.action_ref }}
+      uses: actions/checkout@v4.1.1
+      with:
+        repository: ${{ env.action_repo }}
+        ref: ${{ env.action_ref }}
```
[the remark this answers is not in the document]
iainlane: [see docs](https://docs.github.com/en/actions/learn-github-actions/contexts)

> Do not use in the run context. To make this context work with composite actions, reference it within the env context of the composite action.

actions/trigger-argo-workflow/action.yaml
```diff
@@ -0,0 +1,137 @@
...
+        version: "tags/v3.5.1"
+        file: "argo-linux-amd64.gz"
+        target: ${{ github.workspace }}/bin/argo.gz
+      if: steps.restore.outputs.cache-hit != 'true'
+
+    - name: gunzip
+      id: gunzip
+      shell: sh
+      run: |
+        pwd
```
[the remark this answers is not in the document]
iainlane: can do

actions/trigger-argo-workflow/README.md
```diff
@@ -0,0 +1,49 @@
+# Trigger Argo Workflow Action
+
+This GitHub action triggers an Argo workflow in one of the Grafana Labs Argo
+Workflows instances. It contains a small wrapper around the `argo` CLI, which is
+downloaded as part of the action.
+
+As our Argo Workflows instances require authentication, this workflow will only
+work when triggered from a `grafana`-owned repository. We would welcome
+contributions to extend this action to work with any instance, or you are free
+to fork and modify to run on your own instances.
```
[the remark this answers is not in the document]
iainlane: I don't think we should put work into that right now. If someone else wants to, they can.

Do you have a suggestion here? Want me to add the link or something?

actions/trigger-argo-workflow/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## How to use
+
+## Inputs
+
+- `instance`: The instance to use (`dev` or `ops`). Defaults to `ops`.
+- `namespace`: Required. The namespace to trigger the workflow in.
+- `command`: The command to run. Defaults to `submit`.
```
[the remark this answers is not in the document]
iainlane: What do you want me to do? I wanted to support people running anything they want, but if you want I can remove that.

actions/trigger-argo-workflow/action.yaml
```diff
@@ -0,0 +1,136 @@
...
+      with:
+        path: ${{ github.workspace }}/bin
+        key: argo-linux-amd64-3.5.1
+
+    - name: Fetch Github Release Asset
+      id: fetch_asset
+      uses: dsaltares/fetch-gh-release-asset@1.1.1
+      with:
+        repo: "argoproj/argo-workflows"
+        version: "tags/v3.5.1"
```
dsotirakis, line 75: I wonder if we should make the version an actual variable since it's referenced in line 68 as well, in case we want to update it and we don't want to miss a thing.

actions/trigger-argo-workflow/action.yaml
```diff
@@ -0,0 +1,136 @@
...
+        file: "argo-linux-amd64.gz"
+        target: ${{ github.workspace }}/bin/argo.gz
+      if: steps.restore.outputs.cache-hit != 'true'
```
dsotirakis, line 78: Nit, but just for visibility could we move this above the `with` block?

actions/trigger-argo-workflow/action.yaml
```diff
@@ -0,0 +1,136 @@
...
+          fi
+
+          parameters+=("--parameter" "$line")
+        done <<< "${{ inputs.parameters }}"
+
+        if [ -n "${{ inputs.workflow_template }}" ]; then
+          parameters+=("--workflow-template" "${{ inputs.workflow_template }}")
+        fi
+
+        go run github.com/grafana/shared-workflows/actions/trigger-argo-workflow/cmd/trigger-argo-workflow \
```
dsotirakis, line 128: Do we need the full path here (`github.com/grafana/shared-workflows...`) or can we just run from the local path?

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
```diff
@@ -0,0 +1,151 @@
...
+var nameRe = regexp.MustCompile(`^Name:\s+(.+)`)
+
+func (a App) outputWithURI(input *bytes.Buffer) (string, string) {
+	output := input.String()
+
+	matches := nameRe.FindStringSubmatch(output)
+
+	var uri string
+	if len(matches) == 2 {
+		uri = fmt.Sprintf("https://%s/workflows/%s/%s", a.server(), a.namespace, matches[1])
```
dsotirakis, line 66: What happens in case that's not true? Should we throw an error and fail?

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
```diff
@@ -0,0 +1,213 @@
...
+	case "error":
+		return slog.LevelError, nil
+	default:
+		return slog.LevelInfo, fmt.Errorf("invalid log level: `%s`. choose from: debug, info, warn, error", level)
+	}
+}
+
+func main() {
+	// If we're on a terminal, we use tint, otherwise if we're on GitHub actions
+	// we use `willabites/actionslog` to log proepr Actions messages, otherwise
```
dsotirakis, line 44:
```suggestion
	// we use `willabites/actionslog` to log proper Actions messages, otherwise
```

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
```diff
@@ -0,0 +1,213 @@
...
+
+				lv.Set(lvl)
+
+				return nil
+			},
+		},
+		&cli.StringFlag{
+			Name:    flagInstance,
+			EnvVars: []string{"INSTANCE"},
+			Value:   "ops",
```
dsotirakis, line 114: Hm, shouldn't be safer if the default was `dev` instead?

dsotirakis (commented):
A few comments, nits most of them! 
Also I guess it would be nice to have some tests for the `argo.go` file as well?
(I know it can be hard for some functions, but maybe we could try and test some logic there)

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
iainlane: You won't get a URL in the output from the workflow, but it only matters when the `argo` command managed to run successfully (otherwise _that_ would cause the program to fail).

At the time I didn't think that should cause a fatal error - make the GHA run fail - but do you think differently? Would be happy to change it.

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
iainlane: It's about the ease of use. The normal case will be for real CD workloads and I don't want to have a parameter which everybody ends up setting to the same non-default value. I want these actions to be as easy to use as possible.

I'll close this thread, but feel free to push back and reopen if you want!

actions/trigger-argo-workflow/action.yaml
iainlane: Okay, you can't have real variables like that, but you can do a similar thing with "setup" step that sets an env var, let me do that.

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
dsotirakis: Ah cool, that's only for the output. Then no need to cause a fatal error as you say, however it'd may be a good solution to throw a warning or something, like "couldn't get the URI" so we can know where to look in case there's no output. But no strong opinions on that.

actions/trigger-argo-workflow/action.yaml
iainlane: Works, I've done it in 001829f6a0f4cfde04b1e4e580b38338b8d380cf and you can see a run [here](https://github.com/grafana/shared-workflows/actions/runs/7208163232/job/19636503883). Don't worry that it failed - that's because I undeployed the resources this requires to work. We'll bring those back soon. Just notice that the binary was fetched properly 👍

actions/trigger-argo-workflow/action.yaml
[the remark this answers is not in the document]
iainlane: Added a comment to help: 5c79fa59d64e9c882a3b5fae3380ecbb94b8891f

iainlane:
> Also I guess it would be nice to have some tests for the `argo.go` file as well?

I tried a little bit but I ended up abstracting too much under and interface to make it testable which didn't feel very good. I'll give it another go though, thanks for the nudge.

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
iainlane: Alrighty, let me add a warning, good idea

iainlane: Fixed in d7bc9812733d7b981fa9207a81b15e55faf64a2a!

iainlane (commented):
> > Also I guess it would be nice to have some tests for the `argo.go` file as well?

> I tried a little bit but I ended up abstracting too much under and interface to make it testable which didn't feel very good. I'll give it another go though, thanks for the nudge. 

Alright then. I've added some tests in 8eaad20bcc490f51a04b5311833962f805f1315f. I could go deeper (for example I could make an interface so we can swap out running the command itself) but I was getting the feeling I would be getting smaller returns at that point. Let me know what you think!

actions/trigger-argo-workflow/README.md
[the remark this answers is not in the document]
zzehring: I think our 2 short-term options are:
* make at least the action singularly scoped to submit workflows; then it would match the readme description and name of this whole action
* update the name and description of the action to reflect that this action is generic and able to run any argo workflows CLI command. right now, the description, `This GitHub action triggers an Argo workflow in one of the Grafana Labs Argo
Workflows instances.` and name, `trigger-argo-workflow`, suggest to me that the action is only able to submit workflows.

I think both are valid options. Slightly leaning to just update the description and name since you've already configured the wrapper tool to run any arbitrary command. I just think the way we're describing the tool isn't reflective of what it can do.

actions/trigger-argo-workflow/README.md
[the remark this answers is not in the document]
zzehring: Ah sorry. This was more of a note. I don't think we need to do anything right now (at most a comment/TODO/issue?).

actions/trigger-argo-workflow/action.yaml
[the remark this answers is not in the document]
zzehring: Ah thanks for the info!

zzehring (approved):
LGTM!

actions/trigger-argo-workflow/README.md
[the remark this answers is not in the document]
iainlane: 32f367468e20e737743f91ad6890f383f322f34c

iainlane:
Merging this. Thank you for the reviews, they improved the PR a lot!

actions/trigger-argo-workflow/action.yaml
iainlane: IIRC I got some errors about finding files or libraries or something 😅

PR #20, opened 2023-12-14
GAR: Add option to push multiple tags

dsotirakis:
Instead of passing just one docker image, the workflow should be able to push more than one tags for the same image, but also more than one images even if they're different. 

Related PR: https://github.com/grafana/grafana-backstage/pull/148

### DIFF
--- a/actions/push-to-gar-docker/action.yaml
+++ b/actions/push-to-gar-docker/action.yaml
@@ -34,7 +34,6 @@ runs:
       with:
         context: ${{ inputs.build_path }}
         push: ${{ github.event_name == 'push' }}
-        tags: |
-          ${{ inputs.tags }}
+        tags: ${{ inputs.tags }}
         cache-from: type=gha
         cache-to: type=gha,mode=max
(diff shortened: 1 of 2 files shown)


actions/push-to-gar-docker/action.yaml
```diff
@@ -44,6 +35,6 @@ runs:
         context: ${{ inputs.build_path }}
         push: ${{ github.event_name == 'push' }}
         tags: |
-          ${{ inputs.registry }}/${{ inputs.project }}/${{ inputs.repository }}/${{ inputs.image_name }}:${{ inputs.tag }}
+          ${{ inputs.tags }}
```
iainlane, line 38: trivial: Probably can be `tags: ${{ inputs.tags }}` (one line)

iainlane (approved):

actions/push-to-gar-docker/action.yaml
dsotirakis: true!

PR #24, opened 2023-12-29
OIDC: Construct service account based on input

dsotirakis:
Instead of having a hardcoded service account, we should have it constructed based on the attributes of the caller repository. 
This way we can make sure that the caller workflow can only push artifacts to the registry it means to, using the service account which was created for this purpose only.

### DIFF
--- a/actions/push-to-gar-docker/action.yaml
+++ b/actions/push-to-gar-docker/action.yaml
@@ -8,9 +8,11 @@ inputs:
   tags:
     description: |
       List of Docker images to be pushed.
+    required: true
   build_path:
     description: |
       Path to Dockerfile, which is used to build the image.
+    default: "."
   environment:
     description: |
       Environment for pushing artifacts (can be either dev or prod).
@@ -18,6 +20,7 @@ inputs:
   image_name:
     description: |
       Name of the image to be pushed to GAR.
+    required: true
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


actions/push-to-gar-docker/action.yaml
```diff
@@ -4,26 +4,63 @@ inputs:
...
+          exit 1
+        fi
+        echo "project=${PROJECT}" >> ${GITHUB_OUTPUT}
+    - name: Extract metadata (tags, labels) for Docker
+      id: meta
+      uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
+      with:
+        images: "${{ inputs.registry }}/${{ steps.resolve-project.outputs.project }}/${ROOT_REPO}-${{ inputs.environment }}/${{ inputs.image_name }}"
+        tags: ${{ inputs.tags }}
+    - name: Echo tags
```
zerok, line 47: Do you still need this? IIRC docker/metadata-action already provides some useful output here. Or would it perhaps make sense here to use `$GITHUB_STEP_SUMMARY` for this?

zerok (approved):

actions/push-to-gar-docker/action.yaml
dsotirakis: done!

PR #26, opened 2024-01-05
Dockerhub: Add `push-to-dockerhub` action

dsotirakis:
Part of: https://github.com/grafana/deployment_tools/issues/108822

Along with OIDC way of pushing to GAR through GitHub actions, we need a workflow to push Docker images to DockerHub.
We need to make sure that DockerHub will only contain the end-product images (i.e. `grafana/grafana`, `grafana/mimir` etc.) and all the other images which are used for tooling should be pushed to GAR.

The DockerHub credentials are being fetched from vault.

### DIFF
--- a/actions/push-to-dockerhub/action.yaml
+++ b/actions/push-to-dockerhub/action.yaml
@@ -9,7 +9,7 @@ inputs:
     description: |
       List of Docker images to be pushed.
     required: true
-  build_path:
+  context:
     description: |
       Path to Dockerfile, which is used to build the image.
     default: "."
@@ -44,7 +44,7 @@ runs:
     - name: Build and push Docker image
       uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
       with:
-        context: ${{ inputs.build_path }}
+        context: ${{ inputs.context }}
         push: ${{ github.event_name == 'push' }}
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
--- a/actions/push-to-dockerhub/action.yaml
+++ b/actions/push-to-dockerhub/action.yaml
@@ -42,5 +42,5 @@ runs:
         context: .
         file: ./Dockerfile
         push: true
-        tags: ${{ input.tags }}
+        tags: ${{ inputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
--- a/actions/push-to-dockerhub/action.yaml
+++ b/actions/push-to-dockerhub/action.yaml
@@ -36,10 +36,24 @@ runs:
       with:
         images: ${{ inputs.repository }}
 
+    - name: Resolve tags
+      id: resolve-tags
+      shell: bash
+      run: |
+        tags=(${{ inputs.tags }})
+
+        for tag in ${tags[@]}; do
+          image_uri="${{ inputs.repository }}:$tag"
+          image_uris+=("$image_uri")
+        done
+
+        IFS=, image_uris_string="${image_uris[*]}"
+        echo "images=${image_uris_string}" >> ${GITHUB_OUTPUT}
+
     - name: Build and push Docker image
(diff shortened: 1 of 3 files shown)


actions/push-to-dockerhub/action.yaml
```diff
@@ -0,0 +1,63 @@
...
+        username: ${{ env.DOCKERHUB_USERNAME }}
+        password: ${{ env.DOCKERHUB_PASSWORD }}
+
+    - name: Extract metadata (tags, labels) for Docker
+      id: meta
+      uses: docker/metadata-action@9ec57ed1fcdbf14dcef7dfbe97b2010124a938b7
+      with:
+        images: ${{ inputs.repository }}
+
+    - name: Resolve tags
```
zerok, line 43: Do you really need this step? Wouldn't it also work if you pass the tags int othe `docker/metadata-action` and use the output from there?

actions/push-to-dockerhub/action.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - name: Get secrets for DockerHub login
+      id: get-secrets
+      uses: grafana/shared-workflows/actions/get-vault-secrets@main
+      with:
+        common_secrets: |
+          DOCKERHUB_USERNAME=dockerhub:username
+          DOCKERHUB_PASSWORD=dockerhub:password
+
+    - name: Log in to Docker Hub
+      uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a
```
zerok, line 32: Would it be possible to limit the login step somehow to only specific CI events? Like that it's only done when the image would also be pushed?

actions/push-to-dockerhub/action.yaml
dsotirakis: Not sure, since the format they'll come from is:
```
tags: |-
"foo"
"bar"
```

I'll try though.

actions/push-to-dockerhub/action.yaml
dsotirakis: I think this should be specified on the caller, no?

zerok: True, but isn't there something similar down below when setting the `push` property of the "Build and push Docker image" step?

dsotirakis: Hm, I think if we agree that DockerHub will be only used for "prod" images then we can simply fail early if the event is not a push. Then all the "dev" images and images used for tooling etc should live in GAR. But I guess it's going to be some time before this happens 😅 Happy if we leave it as is and create an issue to address that ASAP?

actions/push-to-dockerhub/action.yaml
dsotirakis: Works indeed @zerok, thanks for the suggestion 🙇

actions/push-to-dockerhub/action.yaml
zerok: Absolutely 🙂

actions/push-to-dockerhub/action.yaml
zerok: 🥳 Used that a couple of days ago when rewriting the CI pipeline for one of my projects ^_^

actions/push-to-dockerhub/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+inputs:
+  repository:
+    description: |
+      The caller workflow's repository
+  tags:
+    description: |
+      List of Docker images to be pushed.
+    required: true
+  build_path:
```
zerok, line 12: Sorry, one final nitpick: I think it would be better to name this parameter "context" as that's what it is called Dockerland.

zerok (approved):

actions/push-to-dockerhub/action.yaml
dsotirakis: https://github.com/grafana/shared-workflows/issues/27 @zerok fyi :)

PR #33, opened 2024-01-16
feat: Rename push-to-dockerhub-action

zerok:
This follows the naming of the underlying Docker action and adds a "push" input so that users can control whether the image should be pushed or not.


dsotirakis (approved):
LGTM!

PR #36, opened 2024-01-30
Add `dockerhub-login` composite action

julienduchesne:
To be used for workflows that don't use the same process as the `build-push-to-dockerhub` action


dsotirakis (approved):
Nice!

PR #40, opened 2024-02-02
Adding push-to-gcs action and README

ricky-undeadcoders:


actions/push-to-gcs/action.yaml
```diff
@@ -0,0 +1,61 @@
...
+        echo "repo_name=${REPO_NAME}" >> ${GITHUB_OUTPUT}
+    - name: Get object name and location
+      id: get-object-location
+      shell: sh
+      run: echo "object_path=${{ inputs.destination != '' && inputs.destination || inputs.object }}" >> ${GITHUB_OUTPUT}
+    - name: Construct service account
+      id: construct-service-account
+      shell: sh
+      run: |
+        SERVICE_ACCOUNT="github-${{ steps.get-repository-name.outputs.repo_name }}-${{ inputs.environment }}@grafanalabs-workload-identity.iam.gserviceaccount.com"
```
abannachGrafana, line 48: are service accounts being automatically generated for each repo or is that up to the codeowners to create them in deployment tools?

iainlane: the latter --- we have a TF module internally for that

iainlane:
We need to align this with the internal implementation (the CSP resources we create for each repository). I'm going to this PR as a draft until that side is ready.

actions/push-to-gcs/action.yaml
```diff
@@ -0,0 +1,95 @@
...
+      shell: bash
+      run: |
+        # Construct bucket name
+        if [ "${{ inputs.bucket }}" == "" ]; then
+          BUCKET="grafanalabs-${{ github.repository_id }}-${{ inputs.environment }}"
+        else
+          BUCKET="${{ inputs.bucket }}"
+        fi
+
+        # If glob is set, then we'll use that instead of object
```
dsotirakis, line 60: you mean instead of the path?

actions/push-to-gcs/action.yaml
```diff
@@ -0,0 +1,95 @@
...
+    required: true
+  bucket_path:
+    description: |
+      The path in the bucket to save the object(s). Defaults to the root of the bucket.
+    default: ""
+  environment:
+    description: |
+      Environment for uploading objects (can be either dev or prod).
+    default: dev
+  glob:
```
dsotirakis, line 23: I like this `glob` thing - I wonder if it would make sense to make it part of the `path` so it's more clear to everybody?
Like check if there are `*` or `**` and have a step in this github action which will either resolve to a single object or a glob pattern? No strong opinions, I think it will just make things a little bit easier. Feel free to ignore if you have already thought about it and it works better your way!

actions/push-to-gcs/README.md
```diff
@@ -0,0 +1,84 @@
...
+| ------------- | ------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `path`        | String | (Required) Path to the object(s) to upload. Can be a path to file to upload 1 file, or a folder path to upload that folder and its contents. Can be used in conjunction with the `glob` option. Valid examples include `thing.txt` and `path/to/thing.txt`. Valid examples when also using `glob` include `path/to`. |
+| `bucket`      | String | Name of bucket to upload to. (Default: `grafanalabs-${repository.id}-${environment}`)                                                                                                                                                                                                                                |
```
dsotirakis, line 75: Hm I thought of that as well during the GAR work - I think for bucket names it makes more sense to just use names instead of ids. What I want to say by that is that for example we use `github-<repoID>-<env>@blah` for service accounts, but service accounts are something that won't bother us at all or will only bother us in some extreme/edge cases. In contrast, there will be times when we would like to manually check something from a bucket asap, and it would be more obvious to have something like `grafanalabs-grafana-enterprise-dev` rather than `grafanalabs-123456789-dev`. It's for practical reasons mostly. But still - not strongly opinionated. We can always revisit if discovery is hard!

dsotirakis (commented):
Just a few things - I think it looks good! 💪

iainlane (commented):
There's something that I've been thinking about for a while regarding our actions like these (build + push to GAR [we actually do have what I'm suggesting below for Docker Hub already], now GCS)...

I think we should provide the plain "login" functionality as a separate action, so we offer two things:

1. All in a box like we have here now: push these files for me
2. Get me set up to push and I'll handle the pushing

In other words, what do you think about splitting this into _two_ composite actions:

1. `authenticate-to-gcs-for-repo` which does all the steps apart from `google-github-actions/upload-cloud-storage` and the path manipulating stuff. It would output the bucket name to give to...
2. `push-to-gcs` which calls `authenticate-to-gcs-for-repo` and then `google-github-actions/upload-cloud-storage` after doing any necessary fiddling with paths

Here's the reason: I already know that we have people interacting with GCS from code. They don't need the action to do the pushing for them. We don't support those users when we only provide an opinionated action.

WDYT?

actions/push-to-gcs/action.yaml
ricky-undeadcoders: Ah, good catch. I meant `object`, but I renamed `object` to `path` when I introduced the `glob` option. I will fix.

actions/push-to-gcs/README.md
ricky-undeadcoders: Yeah, I know, I went this route because if you change the name of a GH repo you also have to recreate the bucket. This avoids that; but at this point both you and Iain have requested it... so it might be time to redo it.

I agree the user experience is worse with this route.

ricky-undeadcoders:
> There's something that I've been thinking about for a while regarding our actions like these (build + push to GAR [we actually do have what I'm suggesting below for Docker Hub already], now GCS)...
> 
> I think we should provide the plain "login" functionality as a separate action, so we offer two things:
> 
>     1. All in a box like we have here now: push these files for me
> 
>     2. Get me set up to push and I'll handle the pushing
> 
> 
> In other words, what do you think about splitting this into _two_ composite actions:
> 
>     1. `authenticate-to-gcs-for-repo` which does all the steps apart from `google-github-actions/upload-cloud-storage` and the path manipulating stuff. It would output the bucket name to give to...
> 
>     2. `push-to-gcs` which calls `authenticate-to-gcs-for-repo` and then `google-github-actions/upload-cloud-storage` after doing any necessary fiddling with paths
> 
> 
> Here's the reason: I already know that we have people interacting with GCS from code. They don't need the action to do the pushing for them. We don't support those users when we only provide an opinionated action.
> 
> WDYT?

@iainlane Sounds good to me! I briefly reflected that it had to login 8 times to push my 8 different test cases. I'll get that done.

actions/push-to-gcs/README.md
ricky-undeadcoders: This has been completed.

actions/push-to-gcs/action.yaml
ricky-undeadcoders: I took a stab at implementing your recommendation, and I found it to be more confusing than leaving them separate. It did however give me the opportunity to revisit and simplify by eliminating the logic I was applying to those inputs, so now we simply pass the `path` and `glob` inputs directly to the downstream action. 

In the README I documented several cases for how to use these inputs in conjunction with each other. Let me know what you think!

actions/login-to-gcs/README.md
```diff
@@ -0,0 +1,42 @@
...
+  contents: read
+  id-token: write
+
+jobs:
+  login-to-gcs:
+    name: login-to-gcs
+    steps:
+      - uses: grafana/shared-workflows/actions/login-to-gcs@rwhitaker/push-to-gcs
+        id: login-to-gcs
+```
```
iainlane, line 29: Suggest adding an example of using the bucket here - maybe can show that `gcloud` cli works?

```console
$ gcloud storage cp OBJECT_LOCATION gs://DESTINATION_BUCKET_NAME
```

as in: you don't have to use the `google-github-actions/upload-cloud-storage`

actions/login-to-gcs/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+          BUCKET="${{ inputs.bucket }}"
+        fi
+
+        echo "bucket=${BUCKET}" >> ${GITHUB_OUTPUT}
+    - name: Construct service account
+      id: construct-service-account
+      shell: bash
+      run: |
+        SERVICE_ACCOUNT="github-${{ github.repository_id }}-${{ inputs.environment }}-gcs@grafanalabs-workload-identity.iam.gserviceaccount.com"
+        echo "service_account=${SERVICE_ACCOUNT}" >> ${GITHUB_OUTPUT}
```
iainlane, line 57: nit: these steps could be combined

actions/login-to-gcs/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+      run: |
+        if [ "${{ inputs.environment }}" = "dev" ]; then
+          PROJECT="grafanalabs-dev"
+        elif [ "${{ inputs.environment }}" == "prod" ]; then
+          PROJECT="grafanalabs-global"
+        else
+          echo "Invalid environment. Valid environment variable inputs: dev, prod"
+          exit 1
+        fi
+        echo "project=${PROJECT}" >> ${GITHUB_OUTPUT}
```
iainlane, line 33: also a nit, something that I've found useful:

```suggestion
        echo "project=${PROJECT}" | tee -a ${GITHUB_OUTPUT}
```

(also for the other outputs). It means the variables get logged which can be helpful sometimes.

actions/login-to-gcs/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Resolve GCP project
+      id: resolve-project
+      shell: bash
+      run: |
+        if [ "${{ inputs.environment }}" = "dev" ]; then
+          PROJECT="grafanalabs-dev"
+        elif [ "${{ inputs.environment }}" == "prod" ]; then
```
iainlane, line 27: nit: noticed the two use different comparison operators

```suggestion
        elif [ "${{ inputs.environment }}" = "prod" ]; then
```

actions/login-to-gcs/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+          PROJECT="grafanalabs-global"
+        else
+          echo "Invalid environment. Valid environment variable inputs: dev, prod"
+          exit 1
+        fi
```
iainlane, line 32: suggestion: alternatively this could be written like so

```suggestion
      run: |
        case "${{ inputs.environment }}" in
          dev)
            PROJECT="grafanalabs-dev"
            ;;
          prod)
            PROJECT="grafanalabs-global"
            ;;
          *)
            echo "Invalid environment. Valid environment variable inputs: dev, prod"
            exit 1
            ;;
        esac
```

which might be a bit clearer to read - you decide :slightly_smiling_face:

actions/login-to-gcs/action.yaml
jdbaldry: If we are using bash which https://github.com/grafana/shared-workflows/pull/40/files/ba125891a150ef68376d045749690d376437b07d#diff-8ad545e598c1ecec7544d085676f0f79de19ebdb2dc8309213d1ae2f8144497eR23 implies, we should use the builtin `[[ <TEST> ]]` form and `==`. Only if we are targeting POSIX sh should we use the test `[` binary and `=` equality comparison.

iainlane: why _should_?

jdbaldry: The bash builtin is faster, has more features, and more consistent syntax but ultimately in these cases they work the same. 

I think rules like this that encourage consistency also improve readability.

iainlane: okey dokey. no particular objection - just that if we're going to ask people to abide by something as a rule then let's write it down if not enforce by CI.

we could defer to [the google guide](https://google.github.io/styleguide/shellguide.html) which does cover this case. If you want to make this a requirement, could you raise a PR please?

ricky-undeadcoders: Yep, this was a silly bug. Fixed.

@jdbaldry I'm rather agnostic about how people choose to write their CI scripts (if it works it works), but if you'd like to put together recommendations in a PR we'll be happy to review. Maybe as part of the [GHA user docs](https://github.com/grafana/deployment_tools/blob/master/docs/platform/continuous-integration/github-actions-user-docs.md)?

actions/login-to-gcs/action.yaml
ricky-undeadcoders: I'm happy with how it is because it matches what's in the related `push-to-gar-docker` action. What I would actually like to see here is another shared workflow that both the `gcs` and `gar` can call so we don't have to duplicate code, but I'm afraid that might be a poor design choice given that there's a limit to how many actions can be invoked.

In any case, if we were going to go that route I would like to attack it in another PR.

actions/login-to-gcs/action.yaml
jdbaldry: Yeah, I don't have a problem with "if it works, it works". My drive-by comment was mostly to share knowledge about preferred practice rather than required.

That being said, given the esoteric syntaxes and number of quirks of shell scripting, I think you can get a lot from using tooling to lint and teach. I think shellcheck does a really good job of helping you learn how the shell is working so that you don't think something is working when it isn't.

actions/login-to-gcs/README.md
iainlane: cheers - I would add that as a step in the example, but that's not required :slightly_smiling_face:

actions/login-to-gcs/action.yaml
iainlane: @ricky-undeadcoders  If we're using `==` as recommended, use `[[` and `]]` too to use `bash`'s version of `test`.

actions/login-to-gcs/action.yaml
```diff
@@ -0,0 +1,61 @@
...
+        echo "project=${PROJECT}" | tee -a ${GITHUB_OUTPUT}
+    - name: Get repository name
+      id: get-repository-name
+      shell: bash
+      run: |
```
iainlane, line 37:
```suggestion
```

actions/login-to-gcs/action.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - name: Construct dynamic account vars
+      id: construct-account-vars
+      shell: bash
+      run: |
+        # Construct repo name
+        REPO_NAME=$(echo ${{ github.repository }} | awk -F'/' '{print $2}')
+        echo "repo_name=${REPO_NAME}" | tee -a ${GITHUB_OUTPUT}
+
+        # Construct bucket name
+        if [ "${{ inputs.bucket }}" == "" ]; then
```
iainlane, line 47:
```suggestion
        if [[ "${{ inputs.bucket }}" == "" ]]; then
```

actions/push-to-gcs/action.yaml
```diff
@@ -0,0 +1,86 @@
...
+        repository: ${{ env.action_repo }}
+        ref: ${{ env.action_ref }}
+        path: shared-workflows
+    - name: Resolve GCP project
+      id: resolve-project
+      shell: bash
+      run: |
+        if [ "${{ inputs.environment }}" = "dev" ]; then
+          PROJECT="grafanalabs-dev"
+        elif [ "${{ inputs.environment }}" == "prod" ]; then
```
iainlane, line 54:
```suggestion
        if [[ "${{ inputs.environment }}" == "dev" ]]; then
          PROJECT="grafanalabs-dev"
        elif [[ "${{ inputs.environment }}" == "prod" ]]; then
```

actions/push-to-gcs/action.yaml
```diff
@@ -0,0 +1,86 @@
...
+          PROJECT="grafanalabs-global"
+        else
+          echo "Invalid environment. Valid environment variable inputs: dev, prod"
+          exit 1
+        fi
+        echo "project=${PROJECT}" >> ${GITHUB_OUTPUT}
```
iainlane, line 60:
```suggestion
        echo "project=${PROJECT}" | tee -a ${GITHUB_OUTPUT}
```

actions/push-to-gcs/action.yaml
```diff
@@ -0,0 +1,86 @@
...
+      shell: bash
+      env:
+        bucket: ${{ steps.login-to-gcs.outputs.bucket }}
+      run: |
+        # Join bucket name with bucket_path (Ex: bucket_name/path/to/file)
+        DESTINATION="${{ env.bucket }}/${{ inputs.bucket_path }}"
+        DESTINATION="${DESTINATION%/}"  # strip trailing /
+
+        echo "Setting destination to ${DESTINATION}"
+        echo "destination=${DESTINATION}" >> ${GITHUB_OUTPUT}
```
iainlane, line 78:
```suggestion
        echo "destination=${DESTINATION}" | tee -a ${GITHUB_OUTPUT}
```

for consistency, and if you do this then the `echo` is not needed since `tee -a` is achieving the same thing :slightly_smiling_face:

actions/login-to-gcs/action.yaml
iainlane: righto!

actions/push-to-gcs/README.md
```diff
@@ -0,0 +1,116 @@
...
+name: Upload Files to GCS
+
+on:
+  push:
+    branches:
+      - main
+
+env:
+  ENVIRONMENT: "dev"
+
```
iainlane, line 18:
```suggestion
```

(not needed to add to `with` on the calls since this is the default)

actions/push-to-gcs/README.md
```diff
@@ -0,0 +1,116 @@
...
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  upload-to-gcs:
+    name: upload
+    runs-on: ubuntu-x64-small
```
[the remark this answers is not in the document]
iainlane: ok fair enough!

iainlane (approved):
Is that 404 in CI because this repository doesn't have a GAR? I suspect so - and this PR looks good to me, so feel free to fix that internally, then retry CI and once it passes merge. :+1: 

If that's not it you can dismiss this review

Thanks for all the work!

PR #41, opened 2024-02-02
Add docker file option for actions that build docker images

ricky-undeadcoders:
- Adding file input for docker build actions
- Updating docker/build-push-action to use the same version across actions
- Updating docker context description

### DIFF
--- a/actions/build-push-to-dockerhub/action.yaml
+++ b/actions/build-push-to-dockerhub/action.yaml
@@ -52,7 +52,7 @@ runs:
         tags: ${{ inputs.tags }}
 
     - name: Build and push Docker image
-      uses: docker/build-push-action@v5.0.0
+      uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
       with:
         context: ${{ inputs.context }}
         platforms: ${{ inputs.platforms }}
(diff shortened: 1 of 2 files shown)


actions/build-push-to-dockerhub/action.yaml
```diff
@@ -48,10 +52,11 @@ runs:
         tags: ${{ inputs.tags }}
 
     - name: Build and push Docker image
-      uses: docker/build-push-action@3b5e8027fcad23fda98b2e3ac259d8d67585f671
+      uses: docker/build-push-action@v5.0.0
```
dsotirakis, line 55: @ricky-undeadcoders I think based on [this](https://github.com/grafana/shared-workflows/pull/43) we need to keep the reference here and not the semver

actions/build-push-to-dockerhub/action.yaml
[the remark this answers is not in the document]
ricky-undeadcoders: Done.

actions/build-push-to-dockerhub/action.yaml
ricky-undeadcoders: Done, thanks for linking that.

dsotirakis (approved):

PR #47, opened 2024-02-08
Adding file argument to docker actions

ricky-undeadcoders:
Second attempt to add docker file variable. I went simpler this time and didn't bother trying to combine context and dockerfile. 

I think this is a simpler and cleaner solution that my previous attempt.

### DIFF
--- a/actions/build-push-to-dockerhub/action.yaml
+++ b/actions/build-push-to-dockerhub/action.yaml
@@ -11,7 +11,7 @@ inputs:
     required: true
   context:
     description: |
-      Path to Dockerfile, which is used to build the image.
+      Path to the Docker build context.
     default: "."
   platforms:
     description: |
@@ -21,6 +21,10 @@ inputs:
     description: |
       Push the generated images also to DockerHub
     default: "false"
+  file:
+    description: |
+      The dockerfile to use.
+    default: "Dockerfile"
 
 runs:
   using: composite
@@ -55,3 +59,4 @@ runs:
         push: ${{ inputs.push }}
         tags: ${{ steps.meta.outputs.tags }}
         labels: ${{ steps.meta.outputs.labels }}
+        file: ${{ inputs.file }}
--- a/actions/push-to-gar-docker/action.yaml
+++ b/actions/push-to-gar-docker/action.yaml
@@ -11,7 +11,7 @@ inputs:
     required: true
   context:
     description: |
-      Path to Dockerfile, which is used to build the image.
+      Path to the Docker build context.
     default: "."
   environment:
     description: |
@@ -25,6 +25,10 @@ inputs:
     description: |
       List of arguments necessary for the Docker image to be built.
     default: ""
+  file:
+    description: |
+      The dockerfile to use.
+    default: "Dockerfile"
 
 runs:
   using: composite
@@ -84,3 +88,4 @@ runs:
         tags: ${{ steps.meta.outputs.tags }}
         cache-from: type=gha
         cache-to: type=gha,mode=max
+        file: ${{ inputs.file }}


nafisat2 (approved):
I believe you could implement your original logic if you wanted to using [this](https://docs.github.com/en/actions/learn-github-actions/expressions#example)

dsotirakis (approved):
think that makes more sense since it doesn’t break backwards compatibility. however I was thinking yesterday about a small action to locate the Dockerfile on its own (only if there’s just one of them) so we may not need to do this at all. I will have a think.

ricky-undeadcoders:
> I believe you could implement your original logic if you wanted to using [this](https://docs.github.com/en/actions/learn-github-actions/expressions#example)

I did something similar in [this commit](https://github.com/grafana/shared-workflows/pull/40/commits/4bf7763677e1b7893fac3d82fb94a47c231f1921) (for a different PR), but I like the simpler solution here.

> think that makes more sense since it doesn’t break backwards compatibility. however I was thinking yesterday about a small action to locate the Dockerfile on its own (only if there’s just one of them) so we may not need to do this at all. I will have a think.

Would it would still make sense to include this option for cases where there maybe multiple dockerfiles?

dsotirakis:
@ricky-undeadcoders maybe yes but we can do that as a followup PR

jdbaldry:
Ah, it is because the default for the `docker/build-push-action` is:

> file | String | Path to the Dockerfile. (default {context}/Dockerfile)

https://github.com/docker/build-push-action

jdbaldry:
I expect that any workflow using a context subdirectory and implicit Dockerfile path are broken atm.

I'm happy to update my workflow so it will work but if we want to make this backwards compatible, and fix other workflows in the company, we'd have to revert and try this again.

zerok:
Hm... would it solve the problem if we just removed the default value? As far as I can tell, the upstream action only set the `--file` parameter [if there is a value for it](https://github.com/docker/build-push-action/blob/62d8db0960c97201f2ae1563702b458b5c216a38/src/context.ts#L141).

jdbaldry:
Weird though, I don't see any logic in the docker/build-push-action apart from pushing the `--file` argument if it is present:
1. https://github.com/docker/build-push-action/blob/master/src/context.ts#L57
2. https://github.com/docker/build-push-action/blob/master/src/context.ts#L141-L143

And according to the [Docker build `--file` documentation](https://docs.docker.com/reference/cli/docker/image/build/#:~:text=By%20default%20the,of%20the%20context. behaves like:

> By default the docker build command looks for a Dockerfile at the root of the build context. The -f, --file, option lets you specify the path to an alternative file to use instead. This is useful in cases that use the same set of files for multiple builds. The path must be to a file within the build context. Relative path are interpreted as relative to the root of the context.

Which I understand to mean `Dockerfile` would be used as `"${CONTEXT}/Dockerfile"` but it's definitely not doing that.

The Docker default confirms that it uses `"${PATH}/Dockerfile"` as the default value (where _`PATH`_ is the build context argument) so I guess our default value in this Action needs to be either `"${CONTEXT}/Dockerfile"` or we need to not pass it through if it is not set.

jdbaldry:
I confirmed the behavior on the command line with `docker image build vale --file Dockerfile` (exit 1), and `docker image build vale --file vale/Dockerfile` (exit 0).

jdbaldry:
> Hm... would it solve the problem if we just removed the default value? As far as I can tell, the upstream action only set the `--file` parameter [if there is a value for it](https://github.com/docker/build-push-action/blob/62d8db0960c97201f2ae1563702b458b5c216a38/src/context.ts#L141).

Yeah I would understand that to make this change backwards compatible.

PR #49, opened 2024-02-22
Updating docker docs to include all inputs

ricky-undeadcoders:

PR #50, opened 2024-02-23
fix: Make file argument optional in docker build actions

zerok:
This should fix the backward-compat issue reported in https://github.com/grafana/shared-workflows/pull/47

### DIFF
--- a/actions/build-push-to-dockerhub/action.yaml
+++ b/actions/build-push-to-dockerhub/action.yaml
@@ -24,7 +24,7 @@ inputs:
   file:
     description: |
       The dockerfile to use.
-    default: "Dockerfile"
+    required: false
 
 runs:
   using: composite
--- a/actions/push-to-gar-docker/action.yaml
+++ b/actions/push-to-gar-docker/action.yaml
@@ -28,7 +28,7 @@ inputs:
   file:
     description: |
       The dockerfile to use.
-    default: "Dockerfile"
+    required: false
 
 runs:
   using: composite


zerok:
@jdbaldry Can you please try that fix in a pull-request on your end? :)

zerok:
https://github.com/grafana/writers-toolkit/actions/runs/8016837140/job/21899553135?pr=530 looks good :)

jdbaldry:
I was too slow :D

actions/push-to-gar-docker/action.yaml
```diff
@@ -28,7 +28,7 @@ inputs:
   file:
     description: |
       The dockerfile to use.
-    default: "Dockerfile"
+    required: true
```
iainlane, line 31: doesn't look like this makes it optional?

actions/build-push-to-dockerhub/action.yaml
```diff
@@ -24,7 +24,7 @@ inputs:
   file:
     description: |
       The dockerfile to use.
-    default: "Dockerfile"
+    required: false
```
iainlane, line 27: what happens now there's no default: we pass an empty string as the input to the build-push-action and it uses its default?

actions/push-to-gar-docker/action.yaml
zerok: I'm bad at copy-pasting :D

iainlane (approved):
looks good! let's make sure #49 is updated accordingly

zerok:
> looks good! let's make sure #49 is updated accordingly

Done :)

PR #60, opened 2024-03-18
chore: add bump and release cloud plugin action

abannachGrafana:
Cloud-only plugins follow a very common pattern when it comes to building and releasing

1. Update package.json version
2. Build, test, package, validate
3. Tag
4. Release and push to GCS

- This composite action is a translation of the https://github.com/grafana/grafana-adaptive-metrics-app/blob/main/.github/workflows/release.yml used in almost all the database frontend repositories https://github.com/grafana/gex-plugins outstanding.
- It simplifies the release process by removing the manual steps of updating the `package.json,` `CHANGELOG.md,` and creating a tag.
- By making it a composite action instead of a shared workflow, we can provide the correct tokens and credentials for the action so that it has permission to release and push the plugin.
- Also, by making it an action first, it allows teams to trigger it however they feel necessary.
- *Possible followup*: Create a shared workflow that pulls the details needed from organization secrets. The only caveats would be that the GITHUB_TOKEN used would need to be an Org-installed GitHub App like `databases-frontend-ci-bot` with permissions to commit and bypass to any repo's default branch and a GCS service account with permission to upload to any bucket.


actions/bump-and-release-cloud-plugin/README.md
```diff
@@ -0,0 +1,67 @@
...
+  bump-and-release:
+    runs-on: ubuntu-latest
+    steps:
+      # ... Steps to get necessary secrets
+      - uses: grafana/shared-workflows/actions/bump-and-release-cloud-plugin@main
+        with:
+          github-token: ${{ env.mygithub-token }} # Needs the ability to commit to main and create tags and releases
+          gcs-bucket: ${{ env.GCS_BUCKET }}
+          gcs-service-account-creds: ${{ env.GCS_SERVICE_ACCOUNT_CREDS }}
+          release-level: 'minor' # 'major', 'minor', 'patch', 'prerelease'
```
guicaulada, line 64: Shouldn't this be `inputs.level` to use the release level from the `workflow_dispatch` ?

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,308 @@
...
+      shell: bash
+      run: |
+        echo "GOPRIVATE=github.com/grafana/*" >> $GITHUB_ENV
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with:
+        ref: ${{ inputs.default-branch }}
+        token: ${{ inputs.github-token }}
+        fetch-depth: 0
+
+    - name: Setup credentials to access Grafana private repositories
```
guicaulada, line 65: I wonder if this should be an optional step, considering shared-workflows is a public repository, isn't there some parametrization that could be done here to allow people who don't have access to Grafana private repositories to use this to release their plugins?

In case cloud plugins are internal only and different from other plugins this makes sense, in which case this action is expected to not work for members without access to Grafana private repositories.

Should we leave a note on the readme to make this clear if that's the case?

actions/bump-and-release-cloud-plugin/README.md
abannachGrafana: Good catch, I copy pasta'd the release yaml and didn't think of using the inputs.level.

actions/bump-and-release-cloud-plugin/action.yaml
abannachGrafana: Good point, I think we could modify L#58 to instead of `GOPRIVATE=github.com/grafana/*` to be `GOPRIVATE=github.com/${{ github.repo.org }}/*` (might have my context variable wrong will double check) So that it's org agnostic. 

And then update the name on this step to just `Setup credentials to access private repositories` since it's really just slapping the token before github.com.

abannachGrafana: > In case cloud plugins are internal only and different from other plugins this makes sense, in which case this action is expected to not work for members without access to Grafana private repositories.

Not necessarily, but so far to my knowledge cloud plugins have been internal provisioned plugins. That doesn't mean this couldn't build a OSS plugin, right now it just builds the linux binaries and not all the other possible binaries if there is a go backend.

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,308 @@
...
+      uses: softprops/action-gh-release@9d7c94cfd0a1f3ed45544c887983e9fa900f0564 # v2.0.4
+      with:
+        prerelease: ${{ inputs.release-level == 'prerelease' }}
+        generate_release_notes: true
+        files: |
+          ${{ runner.temp }}/plugin-artifacts/**/*.zip?(.md5)
+        tag_name: ${{ steps.output-version.outputs.version-tag }}
+        token: ${{ inputs.github-token }}
+
+    - id: auth
```
zerok, line 296: I think in such a setup going with OIDC is the preferred approach in the long run, similar to what has been done here: https://github.com/grafana/shared-workflows/pull/40/files#diff-b9fbde687980aa46fe36ddbd7d7ac35be4e057ab7f6e6a102e269261fbbe8a52R44

But since the credentials are "just" an input, this should be ok in this context.

@dsotirakis WDYT?

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,308 @@
...
+      uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+      with:
+        node-version: lts/iron
+        check-latest: true
+        cache: ${{ inputs.package-manager }}
+
+    - name: Setup Go environment
+      uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+      with:
+        go-version: '1.21'
```
zerok, line 79: Is this version something that could vary from plugin to plugin or that could be extracted from the context where the Action is used?

actions/bump-and-release-cloud-plugin/action.yaml
dsotirakis: Hm, totally agree with @zerok. @abannachGrafana maybe it's better to do this now as part of this PR. We would like to move away from SA secrets and GitHub tokens in general since we have the OIDC auth setup already in place.

abannachGrafana: > move away from SA secrets and GitHub tokens

I'm on board. 

Are service accounts being automatically generated for each repo or is that up to the codeowners to create them in deployment tools?

Since the credentials are inputs right now that opens up to anybody being able to leverage this action. Assuming we go the OIDC route, does that restrict this action to Grafana org use only? Happy to make the change if it simplifies workflows from having to gather creds themselves, but just want to make sure I understand the implications.

actions/bump-and-release-cloud-plugin/action.yaml
abannachGrafana: Wow https://github.com/grafana/plugin-validator has come a long way since I looked at the repo. This should totally be customizable!

actions/bump-and-release-cloud-plugin/action.yaml
dsotirakis: > Assuming we go the OIDC route, does that restrict this action to Grafana org use only?

yep!

dsotirakis: we still don't have a mechanism to auth and push to GCS though, we are working on it atm (cc @ricky-undeadcoders)
what we can do for now is to create an SA with write perms to `${{ inputs.gcs-bucket }}` until we fix the issue. I'll create a PR.

actions/bump-and-release-cloud-plugin/README.md
```diff
@@ -0,0 +1,70 @@
...
+jobs:
+  bump-and-release:
+    runs-on: ubuntu-latest
+    steps:
+      # ... Steps to get necessary secrets
+      - uses: grafana/shared-workflows/actions/bump-and-release-cloud-plugin@main
+        with:
+          github-token: ${{ env.mygithub-token }} # Needs permission to commit to main and create tags and releases
+          gcs-bucket: ${{ env.GCS_BUCKET }}
+          gcs-service-account-creds: ${{ env.GCS_SERVICE_ACCOUNT_CREDS }}
```
dsotirakis, line 66: Making a comment here so we don't forget to remove this after OIDC is in place!

actions/bump-and-release-cloud-plugin/action.yaml
abannachGrafana: > > Assuming we go the OIDC route, does that restrict this action to Grafana org use only?
> 
> yep!

I don't have a horse in either race. So as long as we're good with that I'm happy with the decision. After @ricky-undeadcoders PR is merged we can update this to leverage it 👍

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,320 @@
...
+      uses: google-github-actions/auth@55bd3a7c6e2ae7cf1877fd1ccb9d54c0503c457c # v2.1.2
+      with:
+        credentials_json: '${{ inputs.gcs-service-account-creds }}'
+
+    - id: upload-to-gcs
+      name: Upload assets
+      uses: google-github-actions/upload-cloud-storage@22121cd842b0d185e042e28d969925b538c33d77 # v2.1.0
+      with:
+        path: ${{ runner.temp }}/plugin-artifacts
+        destination: ${{ inputs.gcs-bucket }}
```
iainlane, line 319: is this a central bucket for every plugin, or are we using one for each plugin's repo?

abannachGrafana: Depends on the repo's setup, but my team has a bucket per plugin. I think grafana owned catalog plugins are stored in a centralized bucket though.

actions/bump-and-release-cloud-plugin/README.md
dsotirakis: hey @abannachGrafana 👋 
based on some recent discussion we believe that this PR should fit better in `grafana/grafana-github-actions` repo since it’s Grafana specific. we haven’t communicated this to people yet but this is where we are heading. wdyt?

abannachGrafana: I'm fine with putting it there. I was more wondering about the OIDC portion that you were working on in deployment tools and what if any steps I need to take to make that happen in the action.

jarben:
It's great to see this PR @abannachGrafana! Any plans to add support for [grafana/plugin-e2e](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/ci)?

Also +1 to use the shared org crendetials

abannachGrafana:
> It's great to see this PR @abannachGrafana! Any plans to add support for [grafana/plugin-e2e](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/ci)?
> 
> Also +1 to use the shared org crendetials

I haven't looking into `grafana/plugin-e2e` yet, but I don't see why not. In Adaptive Metrics we have a slightly different approach to e2e https://github.com/grafana/grafana-adaptive-metrics-app/blob/94928ee88a98788ed8830e264cfc1f26d107d8bd/.github/workflows/ci.yml#L75-L183 Which actually builds a self contained docker image of Grafana to be used as a service container in the actual test run. This resolved a few issues we'd run into where login wasn't available due to the image not being fully ready. Also if the tests fail we can pull the image locally and test exactly as it was tested in the pipeline. Which is a frustration point when "It works on my machine!" sometimes the build is just slightly different in the pipeline.

That all being said. I'm open to incorporating e2e how it's done with grafana/plugin-e2e. I'd make it conditional like the changelog updates.

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,320 @@
...
+      shell: bash
+      run: |
+        echo "GOPRIVATE=github.com/${{ github.repository_owner }}/*" >> $GITHUB_ENV
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with:
+        ref: ${{ inputs.default-branch }}
+        token: ${{ inputs.github-token }}
+        fetch-depth: 0
+
+    - name: Setup credentials to access private repositories
```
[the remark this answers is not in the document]
yduartep: +1

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,320 @@
...
+      shell: bash
+      run: |
+        yarn version --no-git-tag-version --${{ inputs.release-level }}
+
+    - name: Bump version
+      id: bump-version
+      if: inputs.package-manager != 'yarn' || steps.is-modern-yarn.outputs.modern == 'true'
+      shell: bash
+      run: |
+        ${{ inputs.package-manager }} version ${{ inputs.release-level }}
```
yduartep, line 138: this can be optional? for example in our plugin (grafana-aws-app and cloud-home-app) we bump the version in package.json and update the Changelog in a previous step where we automatically create a PR with both changes to be approved by other team member and then, when this PR is merged, we push the tag to start the release process. 

Our Changelog is updated automatically based on the PR titles merged using conventional commits, with specific format and includes some specific changes not all the things unreleased. For that reason I prefer to have bump and update Changelog optional both.

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,320 @@
...
+        # We want to stage files now so we don't have to worry about generated files later
+        git add .
+
+    - name: Lint frontend
+      shell: bash
+      run: ${{ inputs.package-manager }} run lint
+
+    - name: Test frontend
+      shell: bash
+      run: ${{ inputs.package-manager }} run test:ci
```
yduartep, line 167: What about including optionally e2e tests execution?. There are some plugins that are already running playwright tests and other planning to use the new @grafana/plugin-e2e tool. Maybe include another input parameter?

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,320 @@
...
+        token: ${{ inputs.github-token }}
+        commit-message: 'chore(version): bump to version [${{ steps.output-version.outputs.version }}] (${{ inputs.release-level }}) [BOT]'
+
+    - name: Create tag
+      uses: rickstaa/action-create-tag@a1c7777fcb2fee4f19b0f283ba888afa11678b72 # v1.7.2
+      with:
+        commit_sha: ${{ steps.commit-version-bump.outputs.commit-sha }}
+        github_token: ${{ inputs.github-token }}
+        message: "chore(tag): tag version ${{ steps.output-version.outputs.version-tag }} [BOT]"
+        tag: ${{ steps.output-version.outputs.version-tag }}
```
yduartep, line 285: if a plugin want to do bump and tag in 2 separate steps I would trigger my release creating the tag manually but this action forces me to do the release on your way all in one with not too much flexibility

actions/bump-and-release-cloud-plugin/action.yaml
```diff
@@ -0,0 +1,320 @@
...
+      with:
+        credentials_json: '${{ inputs.gcs-service-account-creds }}'
+
+    - id: upload-to-gcs
+      name: Upload assets
+      uses: google-github-actions/upload-cloud-storage@22121cd842b0d185e042e28d969925b538c33d77 # v2.1.0
+      with:
+        path: ${{ runner.temp }}/plugin-artifacts
+        destination: ${{ inputs.gcs-bucket }}
+        parent: false
```
yduartep, line 320: Is it possible to add new steps to optionally create deployment_tools PRs to bump version in dev | ops | prod ... env automatically? [this is an example](https://github.com/grafana/grafana-aws-app/blob/5c04bc1fd3b3604b9e0e03e9fffe52cb62a0925a/.github/workflows/release.yml#L189)

actions/bump-and-release-cloud-plugin/action.yaml
abannachGrafana: The intention of this action is to do that bump,build,test,release.

What would your goal of using the action be at a higher level? Just the ability to build and test the plugin? or a simpler way to cut a GitHub release and push to GCS?

yduartep: both but not including the version and Changelog update in the same action as mandatory

iainlane:
Folks, I think things have moved on both for this repo and in general since the activity was seen back in March/April/May. I'm making a similar comment to https://github.com/grafana/shared-workflows/pull/71#issuecomment-2374115522. This action would be better served by living in a repo closer / common to the cloud plugins themselves.

Nothing's being lost - the code is still here, it would just need to be moved to a more appropriate home. 🙂 

Thanks!

PR #65, opened 2024-03-20
docs: update pr linter docs

fcjack:
The config file in the docs is missing the module.exports and makes the action to fail.
I am updating the docs to have the default example that works with the action.


jarben (approved):
LGTM!

PR #71, opened 2024-03-22
Add bundle plugin

SandersAaronD:
This bundles a grafana plugin, which can have a backend or no backend, into a properly signed folder of zip files for both current version and "latest" which can be uploaded in order to publish the plugin. By default it zips from `./dist` into `__to-upload__`, but both of those are configurable parameters.

This workflow is structured the way it is with most logic in a sub-package to facilitate using the package as a standalone command outside of a github action if someone has a workflow that calls for that or is using this logic somewhere outside of Grafana to package a plugin.

Implemented in typescript for good portability and maintainability and to integrate well with github actions without introducing any extra dependencies (on, e.g., golang runtimes or binaries). The same thing can presently be accomplished by downloading a grabpl binary onto a machine with a golang runtime and calling it, as can be seen [here](https://github.com/grafana/grafana/blob/7febba947e97aa659f2680b9817ade5ad13fce0b/.github/workflows/core-plugins-build-and-release.yml).


dsotirakis:
Hey @SandersAaronD 👋 
Do you have any updates on this about getting it moved from this repo?

guicaulada:
Still waiting on update from @SandersAaronD

iainlane:
Hey folks, we've had discussions here and in other channels that this PR is considered to specialised to be included here.

As a result of this PR we [documented](https://github.com/grafana/shared-workflows/blob/main/CONTRIBUTING.md#scope-of-shared-workflows) what the scope of this repository is, which was something we lacked before.

I'm going to close this PR. I encourage you @SandersAaronD to find a home closer to the repos that will need this.

PR #73, opened 2024-03-25
Add action for publishing a plugin with a backend on grafana.com

SandersAaronD:
Based on action in grafana/grafana [here](https://github.com/grafana/grafana/blob/main/.github/workflows/core-plugins-build-and-release.yml#L203-L268):.

Hasn't been tested, I need to make sure I have gcom api root right first.


SandersAaronD:
> We're having a couple of internal discussions about exactly what the scope of this repo should be. Both internally and we have #74 here too now.
> 
> If it's okay I'll mark this PR as a draft for now, pending the outcome of those discussions. Then we'll either un-draft, or close and move to https://github.com/grafana/grafana-github-actions.

I don't have any strong opinion about scope or how we organize plugins, but I've already moved this logic once and it currently works for my workflow. I am probably unable to spend time putting technical effort into moving it again.

SandersAaronD:
Leaving this open entirely so it's visible, since this branch is still being used in a team workflow.

PR #77, opened 2024-04-04
Updating README to include contributor requirements

ricky-undeadcoders:
Any feedback is welcome. 

Note:
To ensure that that PRs require reviews from the teams that own the updated files, we have the following branch protection rule in place for `main`:
```
Require review from Code Owners: Require an approved review in pull requests including files with a designated code owner. 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,31 @@
 # shared-workflows
 
-A public-facing, centralized place to store reusable GitHub workflows and action
-used by Grafana Labs. See the `actions/` directory for the individual actions
-themselves.
+Welcome to Platform Productivity's shared workflows repository!
+
+This public-facing repository contains a collection of GitHub Actions workflows designed to streamline and automate common tasks across Grafana projects. These shared workflows are meant to be invoked from multiple repositories, promoting consistency and efficiency in our development processes.
+
+Workflows in this repository fall into two categories:
+
+1. **Common Workflows:** These are highly generic actions that any Grafana repository might need in its CI pipeline. Examples include logging into Docker, pushing Docker images, tagging Docker images, logging into AWS, publishing RPMs, and more. These actions are stored in the `actions/` folder and are specifically maintained by the platform-productivity team.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,21 +1,31 @@
...
+
+This public-facing repository contains a collection of GitHub Actions workflows designed to streamline and automate common tasks across Grafana projects. These shared workflows are meant to be invoked from multiple repositories, promoting consistency and efficiency in our development processes.
+
+Workflows in this repository fall into two categories:
+
+1. **Common Workflows:** These are highly generic actions that any Grafana repository might need in its CI pipeline. Examples include logging into Docker, pushing Docker images, tagging Docker images, logging into AWS, publishing RPMs, and more. These actions are stored in the `actions/` folder and are specifically maintained by the platform-productivity team.
+
+2. **Group-Specific Workflows:** These are workflows tailored to specific groups but used by multiple repositories. Examples include tests specific to a project that need to be called from multiple repos or steps to publish a plugin for a repo. Workflows in this category are owned by the team that contributed them and are maintained accordingly. These actions should be stored in a folder named after the group that's responsible for them (Ex: `grafana/`).
```
zerok, line 11: I'm not sure that putting GitHub workflows into folders other than `.github/workflows` actually works. According to https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow not even subfolders are supported 🙁

ricky-undeadcoders: You can! I tested using this [test action](https://github.com/grafana/shared-workflows/tree/rwhitaker/subfolder-test) I wrote and was able to invoke it successfully.

See the screenshot below:
Screenshot 2024-04-18 at 15 27 08

ricky-undeadcoders: Huh, maybe I'm missing something here. This is an action in a subfolder in the `shared-workflows` repository that I successfully invoked from an upstream repo. The action also matches the format of the actions that we already have built in the `actions` folder (`login-to-gar`... etc.). 

Does that not meet the requirements of a reusable workflow?

zerok: For actions that's correct. They can be pretty much everywhere inside a repository.

A reusable workflow is sadly a different best and they have to live directly within .github/workflows. https://github.com/grafana/shared-workflows/blob/main/.github/workflows/publish-techdocs.yaml would be an example for a reusable workflow which has the `workflow_call` trigger 🙂

PR #78, opened 2024-04-05
feat(build-push-to-gar): Expose platforms parameter

andriikushch:
This PR is inspired by: 
- https://github.com/grafana/shared-workflows/tree/main/actions/build-push-to-dockerhub

It exposes the `platforms` parameter and will allow to make builds for multiple platforms/architectures.


iainlane (commented):
could you add this to the `inputs` block at the top too?

andriikushch:
> could you add this to the `inputs` block at the top too?

Good point @iainlane 👍 I have missed it. Now it should be there 😃

iainlane (approved):

iainlane:
thanks!

PR #79, opened 2024-04-05
`push-to-gar-docker`: Add caching options

dsotirakis:
Adds caching options for container cache. Defaults to `type=gha` if no cache is provided.

### DIFF
--- a/actions/push-to-gar-docker/README.md
+++ b/actions/push-to-gar-docker/README.md
@@ -46,3 +46,5 @@ jobs:
 | `build-args`  | String | List of arguments necessary for the Docker image to be built.                        |
 | `file`        | String | Path and filename of the dockerfile to build from. (Default: `{context}/Dockerfile`) |
 | `platforms`   | List   | List of platforms the image should be built for (e.g. `linux/amd64,linux/arm64`)     |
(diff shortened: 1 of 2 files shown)


jdbaldry (commented):
This is a cool feature! 

After reading the blog post I had the question:

> Do we know what kind of error repo maintainers will see if they hit the 10GB per-repository cache limit?

The answer is in https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

 > If you exceed the limit, GitHub will save the new cache but will begin evicting caches until the total size is less than the repository limit. The cache eviction process may cause cache thrashing, where caches are created and deleted at a high frequency.

zerok (approved):

iainlane (commented):
Could we add this to `build-push-to-dockerhub` too? The two similar actions should keep parity with each other as far as we can make them.

dsotirakis:
> Could we add this to `build-push-to-dockerhub` too? The two similar actions should keep parity with each other as far as we can make them.

hmmm yes but also I had this crazy idea - one action to rule them all. resolve the registry name and decide how/where to login. I will give it a thought

PR #83, opened 2024-04-12
push-to-gar-docker: Fix cache inputs

iainlane:
The way they were being validated was wrong. But we don't need a step to resolve them using a fallback value - we can achieve the same with `default` on the input instead.


zerok (approved):

zerok:
Closing my competing PR :) https://github.com/grafana/shared-workflows/pull/82

iainlane:
`spiderman fight gif` :grin:

PR #84, opened 2024-04-12
feat(pub-techdocs): add input for running `actions/checkout`

zzehring:
This option allows users to _not_ run `actions/checkout` of the current repo. This is useful in cases of generating docs in a previous workflow job.

### DIFF
--- a/.github/workflows/publish-techdocs.yaml
+++ b/.github/workflows/publish-techdocs.yaml
@@ -17,6 +17,11 @@ on:
         required: false
         type: string
         default: "."
+      checkout-repo:
+        description: "Whether to perform the `checkout` action at the start of the workflow. Setting to `false` is useful when generating docs in previous job. Default is `true`."
+        required: false
+        type: boolean
+        default: true
 
 jobs:
   generate-and-publish-docs:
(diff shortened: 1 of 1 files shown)


iainlane (approved):
:+1: 

hmm we should have a readme for these too, like we do for `actions/`

zzehring:
> hmm we should have a readme for these too, like we do for `actions/`

Ah good idea. Not sure where we'd want to store the READMEs tho. Individual actions have their own dirs, so makes it relatively straightforward.

WDYT of a `docs/` folder for this repo?

iainlane:
I like that you can see it in the GitHub UI when you browse to the directory, so that's a vote in favour of `README.md`.

But maybe it could be a short summary and a link to more docs elsewhere?

Anyway, that can be another issue :slightly_smiling_face:

zzehring:
Hmmm so are you thinking a big `README` under `.github/workflows` that documents all the workflows in that directory?

iainlane:
Not necessarily, just that when you go to `.github/workflows`, something is on your screen which ends up with you at the documentation. It could be all in one `.md` document with folds to make it not so long, or it could be links somewhere else, whatever's best!

PR #94, opened 2024-04-16
`push-to-gar-docker`: Add `ssh` input

dsotirakis:
Adds `ssh` input as per: https://github.com/docker/build-push-action?tab=readme-ov-file#inputs


iainlane (approved):
:+1: 

For these proxy actions, should we link to the documentation in the upstream action?

dsotirakis:
> 👍
> 
> For these proxy actions, should we link to the documentation in the upstream action?

Yeah I guess so. At least only for the bits and pieces that are the same in both this action and the upstream one. Will add a link next to the `ssh` field as I did with the cache inputs.

Issue #68, opened 2024-03-21
Use relative paths when running actions in this repository


iainlane:
In a lot of places, we have `uses` directives like this

https://github.com/grafana/shared-workflows/blob/45c3ff5c37e485d4c6eb8b4b25603b79df4dc72e/actions/dockerhub-login/action.yaml#L9

These can be problematic if someone uses one of our actions or shared workflows with a pinned version instead of tracking `main`. They will get the latest version of the step instead of whatever was current at their sha, and there's no guarantee that this version will be compatible.

In `trigger-argo-workflow` we're doing this already:

https://github.com/grafana/shared-workflows/blob/45c3ff5c37e485d4c6eb8b4b25603b79df4dc72e/actions/trigger-argo-workflow/action.yaml#L50-L60

and

https://github.com/grafana/shared-workflows/blob/45c3ff5c37e485d4c6eb8b4b25603b79df4dc72e/actions/trigger-argo-workflow/action.yaml#L110

Let's update all the workflows/actions to do that, and document this as the approach.

PR #95, opened 2024-04-17
`GAR`: Use `login-to-gar` in `push-to-gar-docker`

dsotirakis:
Removes duplication - uses one composite action in another.

### DIFF
--- a/actions/push-to-gar-docker/action.yaml
+++ b/actions/push-to-gar-docker/action.yaml
@@ -76,6 +76,12 @@ outputs:
 runs:
   using: composite
   steps:
+    - name: Get repository name
+      id: get-repository-name
+      shell: bash
+      run: |
+        REPO_NAME=$(echo ${{ github.repository }} | awk -F'/' '{print $2}')
+        echo "repo_name=${REPO_NAME}" >> ${GITHUB_OUTPUT}
     - name: Resolve GCP project
       id: resolve-project
       shell: bash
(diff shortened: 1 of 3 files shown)


actions/push-to-gar-docker/action.yaml
```diff
@@ -89,37 +95,17 @@ runs:
...
         fi
         echo "project=${PROJECT}" >> ${GITHUB_OUTPUT}
-    - name: Get repository name
-      id: get-repository-name
-      shell: bash
-      run: |
-        REPO_NAME=$(echo ${{ github.repository }} | awk -F'/' '{print $2}')
-        echo "repo_name=${REPO_NAME}" >> ${GITHUB_OUTPUT}
+    - name: Login to GAR
+      uses: grafana/shared-workflows/actions/login-to-gar@main
```
dsotirakis, line 99: will change this from main to pinned in a followup PR which will resolve https://github.com/grafana/shared-workflows/issues/96

zerok: But when is that pinning then updated? We don't yet have any real versions here.

dsotirakis: yes I was just talking about this to Iain - I guess we should start versioning this thing soon so we can use dependabot for that.

zerok: Agreed 🙂 Until we have that, though, we should refrain from internal pinning IMO.

zerok (approved):

actions/push-to-gar-docker/action.yaml
iainlane: https://github.com/grafana/shared-workflows/issues/68

PR #103, opened 2024-04-24
`push-to-gar-docker`: Remove default cache values

dsotirakis:
Removes the default values for `cache-to` and `cache-from`

### DIFF
--- a/actions/push-to-gar-docker/action.yaml
+++ b/actions/push-to-gar-docker/action.yaml
@@ -37,12 +37,10 @@ inputs:
     description: |
       Where cache should be fetched from
     required: false
-    default: "type=gha"
   cache-to:
     description: |
       Where cache should be stored to
     required: false
-    default: "type=gha,mode=max"
   ssh:
     desctiption: |
       List of SSH agent socket or keys to expose to the build


iainlane (commented):
I can see the _what_ from the description (but also from the code), but not the _why_ - why is this needed? :slightly_smiling_face:

dsotirakis:
> I can see the _what_ from the description (but also from the code), but not the _why_ - why is this needed? 🙂

Oh yes sorry - it’s because users reported this error:

1. Cache step of our action fails:

 > exporting to GitHub Actions Cache:
------
ERROR: failed to solve: Signing certificate with thumbprint 943DC3545AEC524FF3E4C1FC45CE22256D24C501 not found in store /etc/certificates. See https://aka.ms/actionsDbBackups for possible mitigation.
Error: buildx failed with: ERROR: failed to solve: Signing certificate with thumbprint 943DC3545AEC524FF3E4C1FC45CE22256D24C501 not found in store /etc/certificates. See https://aka.ms/actionsDbBackups for possible mitigation.

I’m wondering how to avoid that, or if I can disable layer caching.

so I guess setting up the cache only if it’s needed is the best way to go here. wdyt @iainlane?

iainlane:
I [did a quick search](https://github.com/search?q=org%3Agrafana+push-to-gar-docker+NOT+repo%3Agrafana%2Fshared-workflows&type=code) and it shows a few repos which are using this action. [Here's a run from a public repo](https://github.com/grafana/explore-logs/actions/runs/8892367036/job/24416275611) which ran today and used the cache successfully. :slightly_smiling_face: 

Since it seems to be working for some repos / most repos / all except one repo, it'd be a shame to lose the benefits of caching for all the rest. Maybe an alternative could be to provide an input to disable the cache? Or doing nothing is an option since AFAIK the user already worked around it by supplying this input themselves and we could wait and see if it's broken elsewhere.

iainlane:
@dsotirakis thoughts on closing this one? would still be good to reproduce the original issue to investigate it

dsotirakis:
@iainlane yeah your points are totally valid let’s close for now!

PR #108, opened 2024-05-01
all: Fix according to editorconfig / markdownlint style

iainlane:
I noticed when writing #107 that my editor wanted to rewrite some of the quotes in the file because they weren't following our `.editorconfig`'s style.

Fix all files according to `prettier` and then run prettier during pushes / pull requests so we don't regress. Add something to the README too, to make people aware of this.


SadFaceSmith (approved):
LGTM

PR #109, opened 2024-05-02
publish-techdocs: Add publish input and fix action checkout

zerok:
Pinning will happen in a follow-up PR

### DIFF
--- a/actions/techdocs-rewrite-relative-links/action.yaml
+++ b/actions/techdocs-rewrite-relative-links/action.yaml
@@ -53,7 +53,7 @@ runs:
         # These seem not to be available when using a local action, though, and
         # so we need to fall back:
         action_repo: ${{ github.action_repository || 'grafana/shared-workflows' }}
-        action_ref: ${{ github.action_ref || steps.workflow-ref.outputs.ref }}
+        action_ref: ${{ github.action_ref || 'main' }}
       uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       with:
         repository: ${{ env.action_repo }}
--- a/actions/techdocs-rewrite-relative-links/action.yaml
+++ b/actions/techdocs-rewrite-relative-links/action.yaml
@@ -12,7 +12,7 @@ inputs:
     description: |
       Execute link rewriting without updating the underlying files
     required: false
-    default: 'false'
+    default: "false"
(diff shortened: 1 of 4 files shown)


actions/techdocs-rewrite-relative-links/action.yaml
```diff
@@ -38,13 +38,13 @@ runs:
 
   steps:
     - name: Checkout
-      if: ${{ !inputs.skip-checkout }}
+      if: ${{ inputs.checkout }}
```
dsotirakis, line 41: Should this be `checkout-repo`?

zerok: Or perhaps `checkout-action-repository`?

dsotirakis: Yep, even better 😁

actions/techdocs-rewrite-relative-links/action.yaml
```diff
@@ -38,13 +38,13 @@ runs:
...
       env:
         # In a composite action, these two need to be indirected via the
         # environment, as per the GitHub actions documentation:
         # https://docs.github.com/en/actions/learn-github-actions/contexts
-        action_repo: ${{ github.action_repository }}
-        action_ref: ${{ github.action_ref }}
+        action_repo: ${{ github.action_repository && github.action_repository || 'grafana/shared-workflows' }}
```
dsotirakis, line 48: Could you please explain this to me? Would it be different than `${{ github.action_repository || 'grafana/shared-workflows' }}`?
I guess we would need this expression in case `grafana/shared-workflows` was also a context value where we didn't know if it's not null/empty? But now since it's hardcoded I guess it's the same? Unless I miss something obvious

zerok: That part is currently experimental. For some reason the conditional checkout does not work (it's always checked out) and I ran into issues with `github.action_repository`. This whole thing is just weird right now and I'm still trying to find out what is going on here :(

zerok: Looks like "action_repository" etc. are only available when you don't run the action as local action. We might get somewhere with `workflow_ref` in this scenario, but that also does not work in the pull-request context 🙁

dsotirakis (approved):

PR #110, opened 2024-05-02
lint-shared-workflows: Also lint action definitions

iainlane:
These have a schema, so we can check if they're correct beyond just being valid YAML. Use the JSON schema from schemastore.org to do this - so we have to convert the YAML files to JSON first. For this, we run in the `yq` container and use `yq`'s conversion feature.

~This is going to fail until #109 is merged~merged - works now

See [the run on the first commit in this PR][run] for an example of a failure.

[run]: https://github.com/grafana/shared-workflows/actions/runs/8923793469/job/24508781530?pr=110

### DIFF
--- a/.github/workflows/lint-shared-workflows.yaml
+++ b/.github/workflows/lint-shared-workflows.yaml
@@ -1,16 +1,131 @@
 on:
   push:
-  pull_request:
+
+permissions:
+  contents: read
+  actions: write
 
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
+      - name: Checkout
+        uses: actions/checkout@0ad4b8fadaa221de15dcec353f45205ec38ea70b # v4.1.4
         with:
           ref: ${{ github.head_ref }}
 
-      - uses: creyD/prettier_action@31355f8eef017f8aeba2e0bc09d8502b13dbbad1 # v4.3.0
+      - name: Lint with Prettier
+        uses: creyD/prettier_action@31355f8eef017f8aeba2e0bc09d8502b13dbbad1 # v4.3.0
         with:
           dry: true
           prettier_options: "--check ."
+
+      # Lint .github/workflows/*
+      - name: Lint workflow files
+        uses: raven-actions/actionlint@789059c543ab20522fb3e7240794e13b0f69ad67 # v1.0.3
+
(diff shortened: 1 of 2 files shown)


iainlane:
Works now

dsotirakis (commented):
Just noticed that the action runs twice for the PR - should we do:
```
on:
  push:
     - main
  pull_request:
```
?

iainlane:
> Just noticed that the action runs twice for the PR - should we do:
> 
> ```
> on:
>   push:
>      - main
>   pull_request:
> ```
> 
> ?

Good point, we should avoid duplicate runs. Maybe I could run it on just `push:`. When I was playing with this it was kind of nice to not have to make a PR. I'll see if that works properly for the PR status, and if not I'll do what you say :slightly_smiling_face:

iainlane:
> Good point, we should avoid duplicate runs. Maybe I could run it on just `push:`. When I was playing with this it was kind of nice to not have to make a PR. I'll see if that works properly for the PR status, and if not I'll do what you say 🙂

Looks like it worked. What do you think?

.github/workflows/lint-shared-workflows.yaml
```diff
@@ -1,16 +1,63 @@
...
+          {
+            echo 'converted_files<<EOF'
+            cat converted_files.txt
+            echo 'EOF'
+          } >> "${GITHUB_OUTPUT}"
+
+      - name: Validate action definitions
+        uses: ScratchAddons/validate-json-action@8f71e0683221310e32661c1b1634399858bde75f
+        with:
+          schema: https://json.schemastore.org/github-action.json
```
zerok, line 62: Would this also work with a local JSONSchema file so that we can download and cache it?

zerok (approved):
Looks great 😄 Just a very optional nitpick

.github/workflows/lint-shared-workflows.yaml
iainlane: Yes, here you go: a66a17fd2b6ef0a5f170bdbcb1ff6afc9c67d183!

  1. [Without cache](https://github.com/grafana/shared-workflows/actions/runs/8988827934/job/24690531924)
  2. [With cache](https://github.com/grafana/shared-workflows/actions/runs/8988874065/job/24690679827)

iainlane: I realised with the previous one it wasn't saving the cache. Check the latest: now we explicitly restore it. More faff was required though: you can't directly overwrite a cache entry so I needed to add a step to first delete if we want to do that...

PR #112, opened 2024-05-02
build-push-to-dockerhub, dockerhub-login: Use local action

iainlane:
Check the `shared-workflows` repo out and use a unique local path for all actions, instead of fetching `main`. This means that we are always internally compatible. Previously we would risk breaking backwards compatibility if someone had pinned a SHA, which is recommended.

We use a unique checkout path because `actions/checkout` removes the repository in a "post" step and we don't want to risk removing a checkout an "outer" action is using, in the case of nesting. Picture this scenario:

1. [outer] check shared-workflows out to `./_shared-workflows`
2. [nested] run a nested workflow which also checks shared-workflows out to the same path
3. [nested] do some work and then finish, running the "Post checkout" step to remove `./_shared-workflows`
4. [outer] try to do something in `./_shared-workflows` but oops, it's gone

We've gotten away with this up until now as we've not had a step trying to use a checkout after such a removal might have happened. But we might not be lucky forever so be safe now.

Add documentation for this.

Closes: #68

### DIFF
--- a/.github/workflows/publish-techdocs.yaml
+++ b/.github/workflows/publish-techdocs.yaml
@@ -56,18 +56,18 @@ jobs:
           repository: grafana/shared-workflows
           # TODO: Replace after merge
           ref: main
-          path: _shared-workflows
+          path: _shared-workflows-publish-techdocs
 
       - name: Rewrite relative links
         if: inputs.rewrite-relative-links || inputs.rewrite-relative-links-dry-run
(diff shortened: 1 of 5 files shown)


iainlane:
Example run (just for one of the actions changed here): https://github.com/grafana/wait-for-github/actions/runs/8928591038/job/24525823184

dsotirakis (approved):
good stuff!

PR #120, opened 2024-05-08
build-push-to-dockerhub: Generate SBOM & attestations

iainlane:
This generates an SBOM and attestations for that SBOM and also build provenance. These are also stored in one of GitHub's SigStore instances, so the provenance of images built using this action can be verified.

We do this by repurposing the `syft-sbom-report` action and calling that as a nested composite action, to encapsulate the generation and pushing work. This wasn't a usable action (no `action.yaml`) anyway. It was already referencing the `anchore/sbom-action` we use here.


dblinkhorn:
Closing since this is quite old now.

PR #121, opened 2024-05-09
Adding contributor's guide

ricky-undeadcoders:


CONTRIBUTING.md
```diff
@@ -0,0 +1,41 @@
+# Contributing to Grafana's Shared Workflows Repository
+
+Welcome to the Shared Workflows repository! This repository contains reusable actions designed to streamline and automate common tasks across multiple departments and external organizations. We appreciate your interest in contributing to our shared efforts.
+
+## Types of Workflows
+
+### 1. Reusable Workflows
+Reusable workflows are located in the `.github/workflows/` directory. These workflows are primarily intended for internal use within this repository. They should be used for automating tasks specific to this repository's needs. Reusable workflows are also required if you intend to use the `on: workflow_call` event for workflow triggering. For more information on reusing workflows and using `on: workflow_call`, refer to the [GitHub documentation](https://docs.github.com/en/actions/using-workflows/reusing-workflows).
```
iainlane, line 8: > These workflows are primarily intended for internal use within this repository. They should be used for automating tasks specific to this repository's needs.

We can publish shared-workflows to be used by other repositories, for example [publish-techdocs].

[publish-techdocs]: https://github.com/grafana/shared-workflows/blob/main/.github/workflows/publish-techdocs.yaml

> Reusable workflows are also required if you intend to use the `on: workflow_call` event for workflow triggering.

I thought that's what defined something as a reusable workflow, that it can be called.

CONTRIBUTING.md
```diff
@@ -0,0 +1,41 @@
...
+
+3. **Create a Branch:** Create a new branch for your contribution. Use a descriptive name related to the changes you're making.
+
+4. **Make Changes:** Make your desired changes to the codebase.
+
+5. **Commit Changes:** Commit your changes with clear and concise commit messages.
+
+6. **Push Changes:** Push your changes to your forked repository.
+
+7. **Submit a Pull Request:** Submit a pull request to the main repository. Ensure that your pull request is detailed and includes information about the changes made.
```
iainlane, line 27: I would add something about checking CI results here.

There's also a CLA requirement which people will be notified about if they need to sign it.

CONTRIBUTING.md
```diff
@@ -0,0 +1,41 @@
+# Contributing to Grafana's Shared Workflows Repository
+
+Welcome to the Shared Workflows repository! This repository contains reusable actions designed to streamline and automate common tasks across multiple departments and external organizations. We appreciate your interest in contributing to our shared efforts.
+
```
iainlane, line 4: One thing it's important to call out is the scope of this repository. How about something like:

```suggestion

## Scope of `shared-workflows`
This repository is for generic workflows and actions which can be used across multiple types of projects. For more specialised use-cases, we encourage the use of project- or area-specific shared workflow repositories.

Examples which would fit in here:

- Push a container image to a Docker registry
- Run security scanners on built artifacts
- Read and write objects to cloud object storage (GCS or S3)

And those which would be better in project- or area-specific shared workflow repositories:

- Implement a release process specific to a subset of projects
- Lint project(s) with your team's opinionated toolset

Get in touch with the maintainers [via an issue] if you're unsure in a particular case.

[issue]: https://github.com/grafana/shared-workflows/issues/new

```

CONTRIBUTING.md
```diff
@@ -0,0 +1,41 @@
...
+## Contribution Guidelines
+
+1. **Fork the Repository:** Start by forking the repository to your GitHub account.
+
+2. **Clone the Fork:** Clone your forked repository to your local machine.
+
+3. **Create a Branch:** Create a new branch for your contribution. Use a descriptive name related to the changes you're making.
+
+4. **Make Changes:** Make your desired changes to the codebase.
```
iainlane, line 21: Something about requiring documentation?

iainlane (commented):
cheers for this, it's important! I've left some comments. And also, the linter is failing - could you take a look at that too please? :slightly_smiling_face:

CONTRIBUTING.md
ricky-undeadcoders: I was trying to explain why one might add their action to the `actions/` instead of `.github/workflows/`, but I might still be a little confused on that topic. What were the reasons for us adding `publish-techdocs` as a reusable workflow, and not, for example, [push-to-gar-docker](https://github.com/grafana/shared-workflows/blob/main/actions/push-to-gar-docker/action.yaml)?

> I thought that's what defined something as a reusable workflow, that it can be called.

Can't I also call the `push-to-gar-docker` action with the `uses` stanza?

CONTRIBUTING.md
ricky-undeadcoders: Added as bullet point 5. WDYT?

CONTRIBUTING.md
ricky-undeadcoders: Updated to include this. WDYT?

```
11. **Contributor License Agreement (CLA):** Before we can accept your pull request, you need to sign [Grafana's CLA](https://grafana.com/docs/grafana/latest/developers/cla/). If you haven't, our CLA assistant prompts you to when you create your pull request.
```

CONTRIBUTING.md
ricky-undeadcoders: Added that in. Thank you!

CONTRIBUTING.md
iainlane: Maybe there's some upstream documentation to delegate to so we don't have to explain this ourselves? Here's my understanding but it might be slightly off ...

Reusable workflow:

  * Called from `uses:` on a **job**
  * Specified by having `workflow_call:` in the `on:` section
  * Can itself have multiple jobs
  * Kind of like an `include` statement for creating _jobs_
  * As it's always separate jobs, and it's not that easy to share assets between different jobs in the same run, I would say these are more recommended when the work is self-contained (simple outputs are ok)

Action:

  * Called from `uses:` on a **step**
  * (for composite actions) is essentially a collection of steps, so kind of like an `include` statement for creating _steps_
  * As runs within your existing job, you can have steps before and after it, so you can share files and env. vars between your steps and the action's steps easily

> Can't I also call the push-to-gar-docker action with the uses stanza?

Not as a reusable workflow like

```yaml
job:
  foo:
    uses: ...push-to-gar-docker...

  bar:
    steps:
      - id: do-things
```

since there's no `workflow_call:` trigger (no trigger at all, since this is a composite action)

instead you use this as an action:

```yaml
job:
  foo:
    steps:
      - id: do-some-stuff
      - id: push
        uses: ...push-to-gar-docker...
      - id: do-some-other-stuff
```

CONTRIBUTING.md
iainlane: Lovely :+1:

CONTRIBUTING.md
iainlane: excellent

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Scope of `shared-workflows`
+This repository is for generic workflows and actions which can be used across multiple types of projects. For more specialised use-cases, we encourage the use of project- or area-specific shared workflow repositories.
+Examples which would fit in here:
+- Push a container image to a Docker registry
+- Run security scanners on built artifacts
+- Read and write objects to cloud object storage (GCS or S3)
+  And those which would be better in project- or area-specific shared workflow repositories:
+- Implement a release process specific to a subset of projects
+- Lint project(s) with your team's opinionated toolset
```
zerok, line 13: If this is team-specific, shouldn't that go into the project-specific repository?

CONTRIBUTING.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Contribution Guidelines
+
+1. **Fork the Repository:** Start by forking the repository to your GitHub account.
+
+2. **Clone the Fork:** Clone your forked repository to your local machine.
+
+3. **Create a Branch:** Create a new branch for your contribution. Use a descriptive name related to the changes you're making.
+
+4. **Make Changes:** Make your desired changes to the codebase.
```
zerok, line 33: Perhaps also something about providing test-cases where possible (keep forgetting that myself all the time)? 🙂

CONTRIBUTING.md
ricky-undeadcoders: Thanks you, and I appreciate the friendly "did you try Googling that" poke :-).

I found a blog that helped explain the differences and included that in the README. WDYT?

CONTRIBUTING.md
ricky-undeadcoders: Good catch; updated. WDYT?

CONTRIBUTING.md
ricky-undeadcoders: Could you give an example?

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Scope of `shared-workflows`
+
+This repository is for generic workflows and actions which can be used across multiple types of projects. For more specialised use-cases, we encourage the use of project- or area-specific shared workflow repositories.
+Examples which would fit in here:
+
+- Push a container image to a Docker registry
+- Run security scanners on built artifacts
+- Read and write objects to cloud object storage (GCS or S3)
+  And those which would be better in project- or area-specific shared workflow repositories:
```
iainlane, line 13: Formatting is off here

```suggestion

And those which would be better in project- or area-specific shared workflow repositories:

```

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+- Implement a release process specific to a subset of projects
+- Lint project(s) with a single (configurable) toolset
```
iainlane, line 15: Was misleading due to the above suggestion, but the team-specific one goes here:

```suggestion
- Lint project(s) with your team's opinionated toolset
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+Welcome to the Shared Workflows repository! This repository contains reusable actions designed to streamline and automate common tasks across multiple departments and external organizations. We appreciate your interest in contributing to our shared efforts.
+
+## Scope of `shared-workflows`
+
+This repository is for generic workflows and actions which can be used across multiple types of projects. For more specialised use-cases, we encourage the use of project- or area-specific shared workflow repositories.
+Examples which would fit in here:
+
+- Push a container image to a Docker registry
+- Run security scanners on built artifacts
+- Read and write objects to cloud object storage (GCS or S3)
```
iainlane, line 12: ...and the generic one can go here

```suggestion
- Read and write objects to cloud object storage (GCS or S3)
- Lint projects with a single (configurable) toolset
```

(I removed the "(s)" since we don't do that elsewhere.)

although, to be honest, I find that wording a bit unclear. I'm not sure we would simply wrap a linter here, since we don't have global coding standards for the `grafana` org (AFAIK). So we would end up exposing all/most of the options from the underlying action, in which case projects might as well use it directly.

So maybe we can leave this one out. Or, since one action we have is "lint PR title", we could say something like

```
- Lint projects according to a widely-agreed specification such as [Commitlint](https://commitlint.js.org/).
```

WDYT? also cc @zerok

iainlane (approved):
just a couple more suggestions - mostly because I noticed the formatting in the scope section is borked.

don't need to re-review after you look at this, feel free to go ahead and merge IMO

THANKS!

CONTRIBUTING.md
ricky-undeadcoders: I like the mention of `widely-agreed specification`... I think it represents the point we want to drive home. 

Added and pushed.

PR #127, opened 2024-05-14
build-push-to-dockerhub: Fix check for logging into Docker hub

iainlane:
There's a truthiness problem here: inputs for composite actions are always strings, so `"false"` (which is the default) is evaluating to `true` for this conditional and we're logging into Docker Hub when we don't need to.

See [run with `main`][fail] with the default `push` value of `"false"`:

image

and [run with this branch][okay]

image

I was trying to type this as a review comment but I was too slow :grin: 

[fail]: https://github.com/grafana/wait-for-github/actions/runs/9076711847/job/24940103015#step:4:254
[okay]: https://github.com/grafana/wait-for-github/actions/runs/9076920608/job/24940688779#step:4:252

### DIFF
--- a/actions/build-push-to-dockerhub/action.yaml
+++ b/actions/build-push-to-dockerhub/action.yaml
@@ -51,7 +51,7 @@ runs:
         path: _shared-workflows-build-push-to-dockerhub
 
     - name: Login to DockerHub
-      if: inputs.push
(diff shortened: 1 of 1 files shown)


Elfo404 (approved):
damn, i noticed and did the correct thing for the first step but forgot to do the same here. thanks for this!

PR #141, opened 2024-06-04
Print env vars

iainlane:


.github/workflows/test-get-vault-secrets.yaml
```diff
@@ -18,6 +18,20 @@ permissions:
   id-token: write
 
 jobs:
+  print-github-context:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Print GitHub context
+        env:
+          GITHUB_CONTEXT: ${{ toJson(github) }}
+        run: echo "$GITHUB_CONTEXT"
```
zerok, line 27: This will also include the current token, right?

iainlane: stop looking at drafts! :grin: 

the answer is "sort of": [it does include it, but it's masked in the output](https://github.com/grafana/shared-workflows/actions/runs/9366715643/job/25784822369)

```
"token": "***".
```

Issue #80, opened 2024-04-12
`build-push-to-dockerhub`: Add caching options


dsotirakis:
Add caching options for `build-push-to-dockerhub` action as done [here](https://github.com/grafana/shared-workflows/pull/79).

PR #143, opened 2024-06-05
update inputs to `build-push-to-dockerhub` action

joeyorlando:
- Adds the following inputs which are needed to migrate [this Drone pipeline in `grafana/oncall`](https://github.com/grafana/oncall/blob/dev/.drone.yml#L129-L135):
  - `build-args`
  - `target`
  - `cache-to`
  - `cache-from`
- Modifies the action to allow `push`ing images to Dockerhub from all GitHub Actions event types **except `pull_request`** (previously the logic here was to _only_ allow `push`ing on a build triggered from a `push` event)

Closes https://github.com/grafana/shared-workflows/issues/80

### DIFF
--- a/actions/build-push-to-dockerhub/action.yaml
+++ b/actions/build-push-to-dockerhub/action.yaml
@@ -48,13 +48,13 @@ inputs:
 runs:
   using: composite
   steps:
-    - name: Check if push is allowed
-      if: ${{ inputs.push == 'true' && github.event_name != 'push' }}
-      shell: sh
-      run: |
-        >&2 echo "Publishing to DockerHub is only allowed on push events."
-        >&2 echo "If you still want to build images without pushing them, set the push input to false."
-        exit 1
+    # - name: Check if push is allowed
+    #   if: ${{ inputs.push == 'true' && github.event_name != 'push' }}
+    #   shell: sh
+    #   run: |
+    #     >&2 echo "Publishing to DockerHub is only allowed on push events."
+    #     >&2 echo "If you still want to build images without pushing them, set the push input to false."
+    #     exit 1
 
     - name: Checkout shared workflows
       env:
--- a/actions/build-push-to-dockerhub/action.yaml
+++ b/actions/build-push-to-dockerhub/action.yaml
@@ -34,6 +34,16 @@ inputs:
     description: |
       Target stage to build
     required: false
+  cache-from:
+    description: |
+      Where cache should be fetched from
+    required: false
+    default: "type=gha"
+  cache-to:
+    description: |
+      Where cache should be stored to
+    required: false
+    default: "type=gha,mode=max"
 
 runs:
   using: composite
@@ -90,3 +100,5 @@ runs:
         file: ${{ inputs.file }}
         build-args: ${{ inputs.build-args }}
         target: ${{ inputs.target }}
+        cache-from: ${{ inputs.cache-from }}
+        cache-to: ${{ inputs.cache-to }}
--- a/actions/build-push-to-dockerhub/action.yaml
+++ b/actions/build-push-to-dockerhub/action.yaml
@@ -48,6 +48,8 @@ inputs:
 runs:
   using: composite
   steps:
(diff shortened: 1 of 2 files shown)


dsotirakis (approved):

dsotirakis:
looks like you are also tackling https://github.com/grafana/shared-workflows/issues/80 @joeyorlando

joeyorlando:
@dsotirakis you mean like updating the PR's description to add the "closes <issue_number" indicator?

dsotirakis:
> Adds `build-args` and `target` inputs which are needed to migrate

yes and also

> Adds build-args and target inputs which are needed to migrate

to include caches as well

joeyorlando:
done!

actions/build-push-to-dockerhub/action.yaml
```diff
@@ -25,17 +25,36 @@ inputs:
...
   steps:
-    - name: Check if push is allowed
-      if: ${{ inputs.push == 'true' && github.event_name != 'push' }}
-      shell: sh
-      run: |
-        >&2 echo "Publishing to DockerHub is only allowed on push events."
-        >&2 echo "If you still want to build images without pushing them, set the push input to false."
-        exit 1
+    # - name: Check if push is allowed
+    #   if: ${{ inputs.push == 'true' && github.event_name != 'push' }}
```
joeyorlando, line 52: @Elfo404 why the logic here to only allow pushing on `push` events? I have a valid use case to push an image on a `workflow_dispatch` event.

Any thoughts on how to modify this in a non-breaking manner, or can I simply remove this logic?

Elfo404: the original issue in https://github.com/grafana/shared-workflows/issues/27 mentioned limiting pushing to docker hub only for push events as a mechanism to prevent "non-prod" images to be pushed there. 

some more context here: https://github.com/grafana/shared-workflows/pull/26#discussion_r1444303741 between @zerok and @dsotirakis. as of "why" i'm not entirely sure, but i guess if that comment is still valid we could also allow jobs triggered by release_publish events to push to DH and it probably wouldn't be a breaking change.
 
opinions @dsotirakis?

dsotirakis: @joeyorlando my 2 cents here - I would say we can just say "allow all pushes that don't come from a pull request". the reason behind that is that we mostly want people to start using our GAR dev repositories for pushing from pull requests really. having said that, I can see the day that we will revisit this and we will start allowing pushes for **all** events, closer than ever 😁

dsotirakis: we can always add a small comment saying why we ideally wouldn't like to push to dockerhub from pull requests and what's the best alternative (push to GAR dev)

dsotirakis (dismissed):

actions/build-push-to-dockerhub/action.yaml
```diff
@@ -25,15 +25,36 @@ inputs:
...
+    default: "type=gha,mode=max"
 
 runs:
   using: composite
   steps:
+    # See this conversation for more context as to why we don't want to allow pushes on pull requests
+    # https://github.com/grafana/shared-workflows/pull/143#discussion_r1628314620
     - name: Check if push is allowed
-      if: ${{ inputs.push == 'true' && github.event_name != 'push' }}
+      if: ${{ inputs.push == 'true' && github.event_name != 'pull_request' }}
```
Elfo404, line 54: did you mean this?
```suggestion
      if: ${{ inputs.push == 'true' && github.event_name == 'pull_request' }}
```

dsotirakis: ooooh that's for the check - looks like @Elfo404 is right. so if push input is true and is also a pull_request we should exit. thanks @Elfo404, dismissing the review for now

joeyorlando: ah yes, you're right! good catch. updated 😄

Elfo404 (approved):
🚢 🇮🇹

PR #145, opened 2024-06-11
Add new action to find a PR for a commit

iainlane:
Sometimes it's useful to be able to do things with a pull request after it gets merged. For example, maybe there's a build & deploy triggered on pushes to main and you want to report that back to the originating PR.

For that, you need to be able to go back from a commit to the PR that it came from. This simple action performs that lookup using GitHub's API. By default it will look up the current commit, so you can call the action with no inputs and it does the right thing in the common case. It also supports looking up other commits and working in other repositories, if a token is supplied.

A simple testsuite is also included to validate the workflow on changes.

### DIFF
--- a/.github/workflows/test-find-pr-for-commit.yml
+++ b/.github/workflows/test-find-pr-for-commit.yml
@@ -1,0 +1,61 @@
+name: Test find-pr-for-commit action
+on:
+  push:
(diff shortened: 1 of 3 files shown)


zerok (approved):
LGTM 🙂

PR #146, opened 2024-06-11
Switch from secrets.GITHUB_TOKEN to github.token for shared workflows

ricky-undeadcoders:


iainlane:
Thanks! It looks plausible - maybe you can hack the `if` conditions so that step runs, to  be sure?

ricky-undeadcoders:
> Thanks! It looks plausible - maybe you can hack the if conditions so that step runs, to be sure?

[lint-shared-workflows](https://github.com/grafana/shared-workflows/actions/runs/9469105802/workflow?pr=146#L91) did run successfully; and I'm wary of running `build-lint-pr` because it opens a PR into `main`. 

What I can do is create a new action that creates a PR into this branch, and verify that it works there.

iainlane (approved):
great 👍 cheers

PR #147, opened 2024-06-12
ci: add workflow that lints shell scripts with ShellCheck

jdbaldry:
I believe that ShellCheck is a great example of how linting can help share knowledge.
Each linting error is [documented](https://www.shellcheck.net/wiki/Home) and the feedback it provides helps you write more understandable scripts with fewer bugs.

Separating shell scripts also helps with consistent formatting, makes modifications easier, and facilitates testing.

We could also consider encouraging a style guide for readability (perhaps starting with https://google.github.io/styleguide/shellguide.html).


dsotirakis:
@jdbaldry the changes in the `translate-secrets.sh` seems to have broken the action tests

jdbaldry:
I think the **Test get-vault-secrets action / test (dev/ops) (pull_request)** failures are due to my PR being in a fork. I don't seem to have write permissions to create a repository branch.

I'm assuming **Lint PR title / lint-pr-title (pull_request) ** is expecting some sort of conventional commit-like message in the PR subject so I'll try that but the error message "Error: subject may not be empty; type may not be empty" isn't very useful to someone new to the repository or to the idea of using conventional commit-like PR subjects.

jdbaldry:
@dsotirakis Sorry for leaving this for so long. I've fixed the tests to verify that the behavior is unchanged after the linting fixes.

jdbaldry:
I believe that any remaining CI failures are because I'm working in a fork. If I'm given write access to the repository, I'm happy to create a new branch and confirm there.

iainlane:
> I believe that any remaining CI failures are because I'm working in a fork. If I'm given write access to the repository, I'm happy to create a new branch and confirm there.

I figure it'll be best if we can fix our CI to at least not fail for forks, so I've made an attempt in https://github.com/grafana/shared-workflows/pull/472.

README.md
```diff
@@ -25,25 +24,24 @@ Or, of course, install it in any other way you want.
...
+For example:
 
 ```yaml
 - uses: action/foo@abcdef0123456789abcdef0123456789 # v1.2.3
 ```
 
 [hardening]: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
 
-### Refer to other `shared-workflows` actions using relative paths
+### Use other `shared-workflows` actions with relative paths
```
iainlane, line 40: Just picking an instance at random here. I like the editing you're doing, but I'd like it more if we could enforce it so that you don't need to pop up every now and again and fix our style. Is there a way we can do this? Or at the least, is there a style guide we can refer to?

As I was writing that paragraph I realised that this pull request contains two changes: adding `shellcheck` and copyediting.

I won't _require_ it but it would be better to split this into two. And then in the other one we can discuss how to keep ourselves on the straight and narrow.

iainlane (approved):
Great!

> We could also consider encouraging a style guide for readability (perhaps starting with [google.github.io/styleguide/shellguide.html](https://google.github.io/styleguide/shellguide.html)).

Sounds good to have a followup with that. It's all great stuff, but I'd particularly appreciate us having some guidelines on when _not_ to use shell scripts, and this guide has a section covering that topic.

README.md
jdbaldry: Opened https://github.com/grafana/shared-workflows/pull/475

PR #148, opened 2024-06-12
dependabot: Enable for `trigger-argo-workflow` and `techdocs-rewrite-relative-links`

iainlane:
These are written in Go, and so we need to make sure that ecosystem is listed in `dependabot.yml` to keep their dependencies up to date.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,23 @@ updates:
       interval: "daily"
     pull-request-branch-name:
       separator: "-"
+
+  - package-ecosystem: "gomod"
+    directory: "/actions/trigger-argo-workflow"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -14,3 +14,13 @@ updates:
       interval: "daily"
     pull-request-branch-name:
       separator: "-"
+
+  - package-ecosystem: "gomod"
+    directory: "/actions/trigger-argo-workflow"
```
zerok, line 19: Also for https://github.com/grafana/shared-workflows/tree/main/actions/techdocs-rewrite-relative-links please :)

iainlane: a suggestion would be nice heh

zerok (approved):
Thank you ❤️

PR #155, opened 2024-06-13
build-push-to-dockerhub, push-to-gar-docker: Document checkout step

iainlane:
These actions don't check the target repository out themselves, because the caller is likely to have done so already, and if they've changed anything before calling us then we would want that to be used for the build.

But we didn't document this in our examples, so do that.


dsotirakis (approved):
yes good one. thanks!

PR #159, opened 2024-06-17
Fix metadata link.

jeschkies:


iainlane:
could you take a look at the CI failure please? maybe switch to reference style links at the same time, so the table doesn't get crazy wide

zerok (approved):

zerok:
Thank you 🙂

PR #160, opened 2024-06-18
Update publish-techdocs.yaml

tolzhabayev:
Not sure how to test this but looks like pinning the cli should work https://github.com/backstage/backstage/issues/25303


tolzhabayev:
Maybe have to also pin 
https://github.com/grafana/shared-workflows/blob/ccd4cc7b07f8675b8dca436554caf408de50c67f/.github/workflows/publish-techdocs.yaml#L90

zzehring:
> Maybe have to also pin
> 
> https://github.com/grafana/shared-workflows/blob/ccd4cc7b07f8675b8dca436554caf408de50c67f/.github/workflows/publish-techdocs.yaml#L90
> 
> like that:
> 
> ` run: python -m pip install mkdocs-techdocs-core==1.3.5`

Yea, I think to be safe let's pin that as well.

zzehring (approved):

PR #163, opened 2024-06-25
publish-techdocs: Workload identity provider and SA email aren't secrets

iainlane:
We don't need to get these from org secrets. Let's hardcode them.

If we can stop considering the bucket name a secret then we can remove the `secrets: inherit` requirement from our callers, which will be safer for them.

### DIFF
--- a/.github/publish-techdocs-testdata/docs/README.md
+++ b/.github/publish-techdocs-testdata/docs/README.md
@@ -1,0 +1,3 @@
+# Test docs
+
+Lovely stuff.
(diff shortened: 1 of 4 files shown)


zzehring (approved):
Ooo I like the test workflow! :shipit:

> If we can stop considering the bucket name a secret then we can remove the secrets: inherit requirement from our callers, which will be safer for them.

I can look into this in a followup.

iainlane:
> Ooo I like the test workflow! :shipit:
> 
> > If we can stop considering the bucket name a secret then we can remove the secrets: inherit requirement from our callers, which will be safer for them.
> 
> I can look into this in a followup.

Okay, great! You also mentioned possibly making these into parameters (with our values as defaults) so it's _more_ reusable. That could be done as a followup too, no objection here!

PR #165, opened 2024-06-27
feat(pub-techdocs): make bucket an input

zzehring:
The name of the bucket doesn't need to be a secret, so make it an input instead with a default.

Also, remove `inherit: secrets` from example, as no secrets are used in the workflow now.


iainlane (approved):
yay!

PR #171, opened 2024-07-03
feat: add argo-lint action

Duologic:
Figured this could be an interesting shared workflow to get my feet wet in this repo.

The tricky bit about this repo is that I'm passing along an artifact that contains the files to lint. I'm not sure if this is possible :shrug:


actions/argo-lint/action.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - name: Add binary to path
+      shell: sh
+      run: |
+        echo "${{ github.workspace }}/bin" >> $GITHUB_PATH
+
+    - name: Download artifact
+      uses: actions/download-artifact@65462800fd760344b1a7b4382951275a0abb4808 # v4.3.3
+      with:
+        name: ${{ inputs.artifact_name }}
+        path: _shared-workflows-argo-lint
```
iainlane, line 67: If this is the same as `trigger-argo-workflow`, could we factor it out? Or merge the two actions somehow? (keeping the existing interface)

iainlane (commented):
I think we should be able to use a path on the system instead of an artifact

actions/argo-lint/action.yaml
Duologic: Done, separate action workflow now.

Also changed the argo-lint action to use a path.

Duologic:
Tested here: https://github.com/grafana/deployment_tools/actions/runs/9791357289/job/27035756038?pr=153431

dsotirakis (approved):
after Iain's comments and fixes, lgtm!

Duologic:
Tested here: https://github.com/grafana/deployment_tools/actions/runs/9792579185/job/27041722079

actions/argo-lint/action.yaml
```diff
@@ -0,0 +1,37 @@
...
+        # https://docs.github.com/en/actions/learn-github-actions/contexts
+        action_repo: ${{ github.action_repository }}
+        action_ref: ${{ github.action_ref }}
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with:
+        repository: ${{ env.action_repo }}
+        ref: ${{ env.action_ref }}
+        # substitute your-action with a unique name (within `shared-repos` for your
+        # action), so if multiple actions check `shared-workflows` out, they don't
+        # overwrite each other
```
iainlane, line 27:
```suggestion
```

actions/trigger-argo-workflow/action.yaml
```diff
@@ -40,12 +40,8 @@ runs:
...
 
   steps:
-    - name: Set argo version
-      id: argo-version
-      shell: sh
-      run: |
-        ARGO_VERSION=3.5.1
-        echo "version=${ARGO_VERSION}" >> ${GITHUB_OUTPUT}
+    - name: Install argo-cli
+      uses: grafana/shared-workflows/actions/install-argo-cli@main
```
iainlane, line 44: this needs the same change

Duologic:
Please have another look at this, I think I've covered all points.

iainlane (approved):
it looks good to me 👍 nice one!

one way to verify would be to trigger a run with `shared-workflows@duologic/argo-lint`, see if it works.

Duologic:
It's running here: https://github.com/grafana/deployment_tools/actions/runs/9888625576

PR #172, opened 2024-07-03
Slack Notification Action

ricky-undeadcoders:
Add action and README for sending slack notifications. Also updated internal PR with instructions pointing to the new README here.


actions/send-slack-notification/action.yaml
```diff
@@ -0,0 +1,48 @@
...
+    description: "The time"
+  thread_ts: # threaded timestamp on the message that was posted when using bot token
+    value: ${{ steps.send-slack-notification.outputs.thread_ts }}
+    description: "The timestamp on the message that was posted into Slack when using bot token"
+  ts: # timestamp of message posted
+    value: ${{ steps.send-slack-notification.outputs.ts }}
+    description: "The timestamp on the message that was posted into Slack when using bot token"
+  channel_id: # Id of the channel. If channel name was used as input we can get ID from output for later use when updating message or posting to thread.
+    value: ${{ steps.send-slack-notification.outputs.channel_id }}
+    description: "The channel id of the message that was posted into Slack when using bot token"
```
iainlane, line 29: Most of the comments repeat the description so I think they can be dropped :-)

iainlane (approved):
lovely!

actions/send-slack-notification/action.yaml
ricky-undeadcoders: Done, also got a bug to rename the action from `send-slack-notification` to `send-slack-message`, since we're in fact sending Slack messages with this.

iainlane: good one, yep 👍

PR #173, opened 2024-07-05
`lint-shared-workflows`: Add `actionlint.yaml`

dsotirakis:
Upon testing [here](https://github.com/grafana/shared-workflows/actions/runs/9805007501/job/27073819964), I found that the lint fails if we don't specify the self-hosted runner names explicitly. 
Based on the error in that workflow, we need to create an actionlint.yaml file, which got generated when I ran:
`actionlint -init-config`. 
This list of labels will prevent this check from failing when we use our self-hosted runners.
We use the path.Match Glob syntax, as indicated [here](https://github.com/rhysd/actionlint/blob/main/docs/config.md).

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -4,7 +4,3 @@ self-hosted-runner:
     - ubuntu-amd64*
     - ubuntu-x64*
     - ubuntu-arm64*
-# Configuration variables in array of strings defined in your repository or
-# organization. `null` means disabling configuration variables check.
-# Empty array means no configuration variable is allowed.
-config-variables: null
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,6 +1,6 @@
 self-hosted-runner:
   # Labels of self-hosted runner in array of strings.
-  labels: 
+  labels:
     - ubuntu-amd64*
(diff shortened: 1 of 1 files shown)


.github/actionlint.yaml
```diff
@@ -0,0 +1,10 @@
+self-hosted-runner:
+  # Labels of self-hosted runner in array of strings.
+  labels:
+    - ubuntu-amd64*
+    - ubuntu-x64*
+    - ubuntu-arm64*
+# Configuration variables in array of strings defined in your repository or
+# organization. `null` means disabling configuration variables check.
+# Empty array means no configuration variable is allowed.
+config-variables: null
```
iainlane, line 10: why is this one being disabled explicitly? according to the docs, this is the default

dsotirakis: not sure - this is automatically generated and I left it like this until we actually need it 😄 
are there any config variables you'd like to check now?

iainlane: nope, I think this could be dropped as it's setting a setting to its default value :upside_down_face:

dsotirakis: removed!

iainlane (approved):
👍

PR #183, opened 2024-07-15
feat: add release-please action

dsotirakis:
This PR is responsible for adding the `googleapis/release-please-action`. This action is used to make a `shared-workflows` repo release and tag. 
Releasing and tagging this repo will help add automations around updating the commit hashes of the teams/individuals who use the workflows.

First tag and release: https://github.com/grafana/shared-workflows/releases/tag/v1.0.0
Changelog: https://github.com/grafana/shared-workflows/blob/main/CHANGELOG.md

Notice that there are 3 tags, `v1`, `v1.0` and `v1.0.0`. This happens because there are users who prefer to use either a major, minor or a patch version of any given workflow. We create all three of them in order to make it able for the user to get all the changes without pinning their versions.

**Note:**

In order for release-please-action to work, the PRs need to follow the [conventional commits specification](https://www.conventionalcommits.org/en/v1.0.0/), thus we now use the `lint-pr-title` action for pull requests in this repo.


.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: googleapis/release-please-action@7987652d64b4581673a76e33ad5e98e3dd56832f # v4.1.3
+        id: release
+        with:
+          release-type: simple
+          skip-github-pull-request: true
+      - uses: actions/checkout@v4
```
iainlane, line 17: SHA

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: googleapis/release-please-action@7987652d64b4581673a76e33ad5e98e3dd56832f # v4.1.3
+        id: release
+        with:
+          release-type: simple
+          skip-github-pull-request: true
```
iainlane, line 16: are you sure that we want to do this? to me, releasing on every push seems a bit much... why wouldn't people use the SHA then? open to arguments 🙂

iainlane (commented):
just looking at the docs (interesting project, haven't seen it before)... should we use a [manifest](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md)? mentions it's for monorepos - maybe we could release the stuff separately then?

.github/workflows/release.yml
dsotirakis: 😢

.github/workflows/release.yml
dsotirakis: yeah was unsure about this as well, I firstly left it there so the development was easier, but maybe we need to enable pull requests and have open PRs (which will be updated on every push) until we decide when to merge them

dsotirakis:
> just looking at the docs (interesting project, haven't seen it before)... should we use a [manifest](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md)? mentions it's for monorepos - maybe we could release the stuff separately then?

wanted to make it as simple as possible to begin with, I think the manifest is great but overpowered 😁 
will have a look anyway.

iainlane:
> > just looking at the docs (interesting project, haven't seen it before)... should we use a [manifest](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md?rgh-link-date=2024-07-16T08%3A45%3A32Z)? mentions it's for monorepos - maybe we could release the stuff separately then?
> 
> wanted to make it as simple as possible to begin with, I think the manifest is great but overpowered 😁 will have a look anyway.

ah yeah, the main thing I'm thinking is that when we make releases it should result in Dependabot PRs to people who use our actions, and creating those when there aren't actual changes to deploy... is not that friendly

.github/workflows/release.yml
iainlane: yeah I'd do that. From reading about that action it's the flow they want you to use... collect up the changes in a PR and then set that free when you're ready. Feels nice, and would be good to try it for a bit 🙂

(even if I think the thing about deleting/replacing tags is 💀... but that's what GitHub Actions wants you to do!)

commitlint.config.js
iainlane, line 1: This looks to be the same as [the default](https://github.com/grafana/shared-workflows/blob/865fbaba550f9c55c28e05b3ff0d7c883ab94bac/actions/lint-pr-title/commitlint.config.js), so can it be dropped?

iainlane:
We could have a play with the single-release thing and add a manifest later if you want?

commitlint.config.js
dsotirakis: Put it there in case we want to change it in the future, but I can default to the other one for now.

iainlane (approved):

README.md
```diff
@@ -67,3 +67,23 @@ will ensure actions in this repo are always used at the same commit. To do this:
   with:
     some-input: some-value
 ```
+
+### Releasing a version of shared-workflows
+
+When working with `shared-workflows`, we need to make sure that we won't break backwards compatibility.
+For this reason we need to provide releasable actions so engineers can review all the incoming changes,
+have they set automated update mechanisms (e.g. `dependabot`, `renovate` etc).
```
zerok, line 75: This part of the sentence is somehow wrong.

dsotirakis: Slightly changed - not sure where it was somehow wrong but could you have another look @zerok?

zerok (approved):

PR #194, opened 2024-07-22
bug: fix logic to include `merge_group` triggers

dsotirakis:
Latest failures on main for the merge queue events ([example](https://github.com/grafana/shared-workflows/actions/runs/10041547269/job/27749873811)) happen because there lint-pr-title action works only for `pull_request` or `pull_request_target` events. 

This PR makes sure that it also works for `merge_group`.


iainlane (commented):
I think what we probably want to do is look into https://github.com/wagoid/commitlint-github-action and make _this_ the required check, rather than the PR title.

The commit message is what we actually care about for `release-please` and not the PR title itself

dsotirakis:
> I think what we probably want to do is look into https://github.com/wagoid/commitlint-github-action and make _this_ the required check, rather than the PR title.
> 
> The commit message is what we actually care about for `release-please` and not the PR title itself

yeah still experimenting, marked it as draft

PR #195, opened 2024-07-22
ci(test-find-pr-for-commit): don't run for merge group events

iainlane:
This action-testing workflow doesn't make sense when run in the merge group, since the commits there aren't associated with PRs any more. It's failing.

We could in the future consider doing work to get the association back, but I don't think we have a need for it right now, so we can stop running the check there instead.


zzehring (approved):

iainlane:
ah fml, I'm going to remove `lint-pr-title` from required checks for now since it's blocking everything. cheers @zzehring!

PR #197, opened 2024-07-23
feat: expose workflow_url variable in send-slack-message action

dsotirakis:
It may be useful to be able to have the full workflow url to post the logs for GHA on slack. This PR creates a new output variable called `workflow_url` which can be used any time the `send-slack-message` is used.


iainlane (commented):
can you share how you imagine this is going to be used?

dsotirakis:
> can you share how you imagine this is going to be used?

bah while trying to explain this I realized that what I did does the thing the other way around 😩

PR #203, opened 2024-07-30
Add OpenAPI client generation action

julienduchesne:
This action will be used to generate OpenAPI clients from the various OpenAPI specs we have in the org.

It will replace the copy-pasted workflows in these repos:

- https://github.com/grafana/slo-openapi-client
- https://github.com/grafana/grafana-com-public-clients

And then, it can be used to expand the client generation feature and make it available to other repos.


julienduchesne:
Can't request reviewers. So @grafana/platform-cat, look at this

actions/generate-openapi-clients/action.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - id: generate
+      shell: bash
+      run: ${GITHUB_ACTION_PATH}/generate.sh
+      env:
+        SPEC_PATH: ${{ inputs.spec-path }}
+        OUTPUT_DIR: ${{ inputs.output-dir }}
+        PACKAGE_NAME: ${{ inputs.package-name }}
+        OPENAPI_GENERATOR_VERSION: ${{ inputs.generator-version }}
+    - id: commit
+      uses: stefanzweifel/git-auto-commit-action@8621497c8c39c72f3e2a999a26b4ca1b5058a842 # v5.0.1
```
iainlane, line 43: Nice action. I didn't know about it and I'll definitely be using it 👍

I noticed on its [docs](https://github.com/stefanzweifel/git-auto-commit-action#checkout-the-correct-branch) that you have to check out the `head_ref` of a PR to be able to push. The action should handle this.

Either update the `README`'s example to do this, or have this action check the repo out again itself to a separate path.

iainlane (commented):
Cheers. Looks like a good candidate for this repo for sure.

Can you add a `CODEOWNERS` entry for an appropriate team so they get tagged for reviews? You'll be better than the default owners, especially for the `mustache` file 🙂

actions/generate-openapi-clients/generate.sh
```diff
@@ -0,0 +1,27 @@
+#! /usr/bin/env bash
+set -eo pipefail
+
+# Download openapi-generator-cli
+wget -nv "https://repo1.maven.org/maven2/org/openapitools/openapi-generator-cli/${OPENAPI_GENERATOR_VERSION}/openapi-generator-cli-${OPENAPI_GENERATOR_VERSION}.jar" -O ./openapi-generator-cli.jar
```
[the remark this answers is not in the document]
julienduchesne: Good idea. Gives me a reason to take the download out of the main script. Done!

actions/generate-openapi-clients/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+permissions:
+  contents: write # Only needed if `commit-changes` is set to true
+
+jobs:
+  build-and-publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+      # Check head_ref if working from a pull request
```
iainlane, line 36:
```suggestion
      # Checkout head_ref if working from a pull request
```

iainlane (approved):
Nice, cheers 👍 

Only other suggestion I had (can be done later or never) would be to comment back on the source PR (if the event is appropriate) saying that a commit was pushed to it

actions/generate-openapi-clients/README.md
jdbaldry: nit on the suggestion: "check out" is the verb, "checkout" is the noun for the subcommand https://git-scm.com/docs/git-checkout

```suggestion
      # Check out head_ref if working from a pull request
```

PR #212, opened 2024-08-13
feat: add setup-conftest action

Duologic:
title says it, follows the same pattern as `install-argo-cli`.

### DIFF
--- a/actions/setup-conftest/action.yaml
+++ b/actions/setup-conftest/action.yaml
@@ -39,7 +39,7 @@ runs:
 
     - name: Save to cache
       id: save
-      if: steps.gunzip.outcome == 'success'
+      if: steps.unpack.outcome == 'success'
       uses: actions/cache/save@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
       with:
         path: ${{ github.workspace }}/bin/conftest
--- a/actions/setup-conftest/action.yaml
+++ b/actions/setup-conftest/action.yaml
@@ -18,6 +18,17 @@ runs:
         path: ${{ github.workspace }}/bin/conftest
         key: conftest-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
 
+    # runner.arch options: X86, X64, ARM, or ARM64.
+    # conftest release: arm64, x86_64, ppc64le, s390x.
+    # If it ain't arm64 or x86_64, it'll fall back to runner.arch.
+    - if: runner.arch == 'X64'
+      shell: sh
+      run: echo "ARCH=x86_64" >> "$GITHUB_ENV"
+
+    - if: runner.arch == 'ARM64'
+      shell: sh
+      run: echo "ARCH=arm64" >> "$GITHUB_ENV"
+
     - name: Fetch Github Release Asset
       id: fetch_asset
       if: steps.cache.outputs.cache-hit != 'true'
@@ -25,7 +36,7 @@ runs:
       with:
         repo: "open-policy-agent/conftest"
         version: "tags/v${{ inputs.version }}"
-        file: "conftest_${{ inputs.version }}_${{ runner.os }}_${{ runner.arch }}.tar.gz"
+        file: "conftest_${{ inputs.version }}_${{ runner.os }}_${{ env.ARCH || runner.arch }}.tar.gz"
         target: ${{ github.workspace }}/bin/conftest.tgz
 
     - name: Unpack tarball
--- a/actions/setup-conftest/action.yaml
+++ b/actions/setup-conftest/action.yaml
@@ -1,5 +1,5 @@
-name: Install conftest
-description: Install conftest
+name: Setup Conftest
+description: Setup conftest and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
 
 inputs:
   version:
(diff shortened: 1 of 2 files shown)


actions/setup-conftest/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+        key: conftest-linux-amd64-${{ inputs.version }}
+
+    - name: Fetch Github Release Asset
+      id: fetch_asset
+      if: steps.restore.outputs.cache-hit != 'true'
+      uses: dsaltares/fetch-gh-release-asset@a40c8b4a0471f9ab81bdf73a010f74cc51476ad4 # 1.1.1
+      with:
+        repo: "open-policy-agent/conftest"
+        version: "tags/v${{ inputs.version }}"
+        file: "conftest_${{ inputs.version }}_Linux_x86_64.tar.gz"
```
iainlane, line 28: can we make this work for other arches and potentially OSes? For ourselves, I would love the option to run on arm64 runners.

```suggestion
        file: "conftest_${{ inputs.version }}_${{ runner.os }}_${{ runner.arch }}.tar.gz"
```

actions/setup-conftest/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+runs:
+  using: composite
+
+  steps:
+    - name: Restore cache
+      id: restore
+      uses: actions/cache/restore@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: ${{ github.workspace }}/bin
+        key: conftest-linux-amd64-${{ inputs.version }}
```
iainlane, line 19: and:

```suggestion
        key: conftest-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
```

actions/setup-conftest/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+runs:
+  using: composite
+
+  steps:
+    - name: Restore cache
+      id: restore
+      uses: actions/cache/restore@0c45773b623bea8c8e75f6c82b208c3cf94ea4f9 # v4.0.2
+      with:
+        path: ${{ github.workspace }}/bin
```
iainlane, line 18: optional: just in case the directory happens to have things in it already e.g. prepopulated on the runner or put there by a previous step?

```suggestion
        path: ${{ github.workspace }}/bin/conftest
```

actions/setup-conftest/action.yaml
```diff
@@ -0,0 +1,51 @@
...
+        mv conftest ${{ github.workspace }}/bin
+        chmod +x ${{ github.workspace }}/bin/conftest
+
+    - name: Save to cache
+      id: save
+      if: steps.gunzip.outcome == 'success'
+      uses: actions/cache/save@13aacd865c20de90d75de3b17ebe84f7a17d57d2 # v4.0.0
+      with:
+        path: ${{ github.workspace }}/bin
+        key: ${{ steps.restore.outputs.cache-primary-key }}
```
iainlane, line 46: AFAICS you could use `actions/cache` instead of the separate `restore` and `save` and get the right behaviour, would save a bit of code.

actions/setup-conftest/action.yaml
```diff
@@ -0,0 +1,51 @@
+name: Install conftest
+description: Install conftest
+
+inputs:
+  version:
+    description: |
+      Version of conftest to install.
+    default: 0.55.0
```
iainlane, line 8: suggestion: could default to `latest` like the `fetch-gh-release-asset` action does, and this could become optional then

iainlane (commented):
Great idea 👍 

Could you add a brief `README.md` along the lines of the others we have in this repo?

actions/setup-conftest/action.yaml
Duologic: Latest would need more plumbing to be able to cache, I'd prefer hardcoding it now.

iainlane: no worries 👍

actions/setup-conftest/action.yaml
iainlane: Ah ok, I checked and it [looks like](https://github.com/open-policy-agent/conftest/blob/56d742d241174bbc790d7b6e6cb3a8632ffff576/.goreleaser.yml#L34) they're using something like `$(dpkg --print-architecture)` but with a special case for amd64.

I forgot but I had done this before. Code from an internal action (which could switch to `setup-conftest` 😁):

```bash
# Following upstream's behaviour:
# https://github.com/open-policy-agent/conftest/blob/56d742d241174bbc790d7b6e6cb3a8632ffff576/.goreleaser.yml#L34C36-L34C42
# So this workflow can run on arm64 or x86_64.
ARCH="$(dpkg --print-architecture)"
if [ "${ARCH}" = "amd64" ]; then
    ARCH="x86_64"
fi
```

Duologic: That'll depend on a debian host... I'm having a look at a quick switch case in the steps.

iainlane: As it's `goreleaser` it's probably `go env GOARCH` rather than `dpkg`, thinking about it

Duologic: Tested on the internal branch, worked. :muscle:

iainlane (approved):
looking good 👍

PR #215, opened 2024-08-15
refactor(setup-argo): refactor following feedback

Duologic:
Refactor this action based on mob session and review feedback in relation to https://github.com/grafana/shared-workflows/pull/212


actions/setup-argo/action.yaml
```diff
@@ -0,0 +1,62 @@
...
+      with:
+        path: ${{ github.workspace }}/bin/argo
+        key: argo-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
+
+    - name: Normalize runner.os to match Argo release
+      shell: sh
+      env:
+        OS: ${{ runner.os }}
+      run: |
+        OS=$(echo "$OS" | tr '[:upper:]' '[:lower:]')
```
Duologic, line 26: @iainlane I can't do the same as on https://github.com/grafana/shared-workflows/pull/216 because I also need to lowercase it for argo.

iainlane: Setting up both the variables in one step with a couple of `case` statements might be the most straightforward

iainlane (approved):
Looks good 👍 

A quick `README` would be nice if you have a sec for that

actions/setup-argo/action.yaml
Duologic: What's wrong with the current script? First lowercase it, then `s/macos/darwin`.

Duologic: Merging so I can keep testing it, we can iterate on the finer details in a follow up.

iainlane: I find it a bit convoluted doing all these steps when you could directly map the values

PR #216, opened 2024-08-15
fix(setup-conftest): normalize runner.os

Duologic:
runner.os maps to `macOS`, the release artifact uses `Darwin` ...

### DIFF
--- a/actions/setup-conftest/action.yaml
+++ b/actions/setup-conftest/action.yaml
@@ -18,13 +18,9 @@ runs:
         path: ${{ github.workspace }}/bin/conftest
         key: conftest-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
 
-    - name: Normalize runner.os to match Conftest release
+    - if: runner.os == 'macOs'
       shell: sh
-      env:
-        OS: ${{ runner.os }}
-      run: |
-        [[ "$OS" == 'macOs' ]] && OS='Darwin'
-        echo "OS=$OS" >> "$GITHUB_ENV"
+      run: echo "OS=Darwin" >> "$GITHUB_ENV"
 
     # runner.arch options: X86, X64, ARM, or ARM64.
     # conftest release: arm64, x86_64, ppc64le, s390x.
--- a/actions/setup-conftest/action.yaml
+++ b/actions/setup-conftest/action.yaml
@@ -18,6 +18,14 @@ runs:
         path: ${{ github.workspace }}/bin/conftest
         key: conftest-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
 
+    - name: Normalize runner.os to match Conftest release
+      shell: sh
+      env:
+        OS: ${{ runner.os }}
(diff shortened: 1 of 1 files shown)


actions/setup-conftest/action.yaml
```diff
@@ -18,6 +18,14 @@ runs:
         path: ${{ github.workspace }}/bin/conftest
         key: conftest-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
 
+    - name: Normalize runner.os to match Conftest release
+      shell: sh
+      env:
+        OS: ${{ runner.os }}
+      run: |
+        [[ "$OS" == 'macOs' ]] && OS='Darwin'
+        echo "OS=$OS" >> "$GITHUB_ENV"
```
iainlane, line 27: can't we do it the same way we do the two below?

```suggestion
      shell: sh
      if: runner.os == 'macOs'
      run: |
        echo "OS=Darwin" >> "$GITHUB_ENV"
```

actions/setup-conftest/action.yaml
```diff
@@ -36,7 +44,7 @@ runs:
       with:
         repo: "open-policy-agent/conftest"
         version: "tags/v${{ inputs.version }}"
-        file: "conftest_${{ inputs.version }}_${{ runner.os }}_${{ env.ARCH || runner.arch }}.tar.gz"
+        file: "conftest_${{ inputs.version }}_${{ env.OS || runner.os }}_${{ env.ARCH || runner.arch }}.tar.gz"
```
iainlane, line 47: If you keep it the way you have it, the fallback shouldn't be needed since you unconditionally set it.
```suggestion
        file: "conftest_${{ inputs.version }}_${{ env.OS }}_${{ env.ARCH || runner.arch }}.tar.gz"
```

In other words, pick one or the other of this and the above suggestion 😁

actions/setup-conftest/action.yaml
```diff
@@ -18,6 +18,14 @@ runs:
         path: ${{ github.workspace }}/bin/conftest
         key: conftest-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
 
+    - name: Normalize runner.os to match Conftest release
+      shell: sh
+      env:
+        OS: ${{ runner.os }}
+      run: |
+        [[ "$OS" == 'macOs' ]] && OS='Darwin'
```
iainlane, line 26: btw the [docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#runner-context) say `macOS` with an upper case `S`. they could be wrong, just pointing it out in case this is a typo

iainlane, line 26: (if you keep this approach)

`[[` is a bashism

```suggestion
        [ "$OS" == 'macOs' ] && OS='Darwin'
```

iainlane (approved):
nice catch 👍

Had a couple of comments but you can fix them up as needed and then merge

PR #220, opened 2024-08-16
fix(trigger-argo-workflow): use the setup-argo workflow

fcjack:
## Why?

Use the setup-argo workflow for trigger-argo-workflow


iainlane (approved):

iainlane:
Can you see if this is still needed after 47a1c7f387daf4ef593b82cb6ac2abca0cd7cf73?

PR #227, opened 2024-08-29
feat(aws-auth): add workflow_ref claim

guicaulada:
This is needed by some workflows, for example in case I want to use a reusable workflow I can already use the `job_workflow_ref` claim, but if I want to also control where the `job_workflow_ref` is triggered from, for example a specific workflow running when certain tags are pushed, I would need the `workflow_ref` to the workflow that triggered the `job_workflow_ref`.


nafisat2 (approved):
Could you also update the README with `workflow_ref`? https://github.com/grafana/shared-workflows/blob/aee5ab9e7b35eed475b97055a7153b45c354a19c/actions/aws-auth/README.md?plain=1#L42

PR #240, opened 2024-09-17
feat(deps): use Renovate to manage Actions dependencies

iainlane:
Renovate supports updating actions versions in composite actions and Dependabot doesn't. That means in our `actions/*/action.yml` files we have a lot of out of date deps.

This is a test to see if it'll work for that case. We disable Dependabot for Actions and let Renovate manage them.

Check [logs for the run on this PR](https://github.com/grafana/shared-workflows/actions/runs/10903114856/job/30256479619?pr=240). They look good to me. If it works well after we merge this we could consider making this a shared workflow and putting the secrets in Vault rather than GH secrets, & then also whether we should enable for other dep types too.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,12 +2,6 @@
 
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    schedule:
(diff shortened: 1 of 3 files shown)


zzehring (approved):
Let's give it a try!

PR #249, opened 2024-09-19
fix(renovate): fix action versions, add more triggers

iainlane:
When referencing actions, we'll do:

```yaml
uses: foo/bar-action@123456789 # v1.2.3
```

to reference by SHA but still say what the version was. Dependabot supports updating these comments when creating bumps, but Renovate doesn't do this when the comment points to the wrong version.

We had two cases of that in `codeql-action` and `vault-action`. Here we fix both.

Also, it's easier to test if we can kick off runs manually and if Renovate triggers when its config or workflow are changed, so add triggers for that.

[preset]: https://docs.renovatebot.com/presets-helpers/#helperspingithubactiondigeststosemver

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -42,7 +42,7 @@ jobs:
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
 
       - name: Initialize CodeQL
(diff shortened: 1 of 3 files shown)


dsotirakis (approved):

iainlane:
@dsotirakis can you check again please? what I thought initially was wrong, and really it was that some of the version comments were incorrect!

iainlane:
oh what sorry, I thought that would not merge until you had re-reviewed 😢

Issue #187, opened 2024-07-16
feat: add a manifest to the release-please action


dsotirakis:
Based on this [documentation here](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md), we should consider whether we should start releasing based on a custom manifest.

PR #412, opened 2024-09-26
ci(release-please): add monorepo config

iainlane:
This will allow us to release each of our actions and reusable workflows individually, which will allow us to evolve them without worrying about breaking users. Doing it separately will avoid users being asked to upgrade unnecessarily.

This required reworking the tagging support, since now multiple components can potentially be released at the same time. We handle this in a GitHub script by parsing the outputs of the release-please action and then creating the appropriate major / major.minor tags.

Closes: #187


.github/workflows/release.yml
```diff
@@ -12,22 +12,90 @@ jobs:
...
+
+            console.log(`Components released: ${components.join(', ')}`);
+
+            for (const component of components) {
+              const tag = releases[`${component}--tag_name`];
+              const sha = releases[`${component}--sha`];
+
+              console.log(`Updating major and minor tags for ${component} to ${sha}`);
+
+              const [name, semver] = rsplit(tag, '-', 1);
```
zzehring, line 96: **question**: would the separator need to be `--`?

Not exactly sure what the `rsplit` would give return here, but based on the python definition of `rsplit`, wouldn't the `name` return `${component}-`?

zzehring (commented):
Really like this change. Just a question on the splitting of the tag to get `name` and `semver`.

.github/workflows/release.yml
iainlane: ah nah, this is a different thing we're splitting on.

`--` is a separator in the names of the outputs which the action creates. _keys_ of the outputs object.

This is something else - the _tag_ the action pushed which is a value associated to one of those keys. That's something like `aws-auth-v1.0.0`. So this `rsplit` would give you `["aws-auth", "v1.0.0"]`. Check logs from a test run [here](https://github.com/grafana/shared-workflows/actions/runs/11049876522/job/30696401275). 👍

zerok (approved):
❤️  and another special thank you for avoiding bash 🤣

iainlane:
> ❤️ and another special thank you for avoiding bash 🤣

Always have a low threshold for moving beyond shell scripts 😁!

.github/workflows/release.yml
zzehring: Ahhhhhhh right. Thanks!

PR #430, opened 2024-09-27
Set up golang before argo setup

SimKev2:
Ran into an issue on arm self-hosted runners that `go` couldn't be called to calculate the os/arch when setting up argo. Setting up golang first should allow that code to function again.


iainlane (changes requested):
Cheers for this. I just want to hold this for a little minute as we expect `go` to be in the default path and it seems like an issue with our self-hosted runners that it's not.

PR #452, opened 2024-10-10
feat(publish-techdocs)!: switch to AWS for publish techdocs

zerok:
Part of https://github.com/grafana/enghub/issues/405

Tested with https://github.com/grafana/enghub/actions/runs/11290282786/job/31401811080

Marked as breaking because it renames/changes some of the inputs as we need AWS-specific semantic here.


iainlane (approved):
Thanks! I think this is fine (assuming it's been tested), but it would be in line with what we do for argo WF if we instead have a "dev or ops" input, and then set the region/role ARN accordingly. Then when we move to `cool new cloud` in 5 years time the interface will stay compatible 🙂

zerok:
Updated the inputs and tested against dev and ops 🙂 Will merge after lunch 😄

PR #458, opened 2024-10-12
fix: call setup-go and then setup-argo

mem:
setup-argo needs go, so call setup-go first.

### DIFF
--- a/actions/trigger-argo-workflow/action.yaml
+++ b/actions/trigger-argo-workflow/action.yaml
@@ -53,9 +53,6 @@ runs:
         ref: ${{ env.action_ref }}
         path: _shared-workflows-trigger-argo-workflow
 
-    - name: Setup argo
-      uses: ./_shared-workflows-trigger-argo-workflow/actions/setup-argo
-
     - name: Setup go
       uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
       with:
@@ -63,6 +60,9 @@ runs:
         cache-dependency-path: |
           actions/trigger-argo-workflow/go.sum
         go-version-file: "_shared-workflows-trigger-argo-workflow/actions/trigger-argo-workflow/go.mod"
+
+    - name: Setup argo
+      uses: ./_shared-workflows-trigger-argo-workflow/actions/setup-argo
 
     - name: Map cluster and vault instance from Argo WF instance
       id: argo-instance


iainlane:
what's the problem that you saw? we'd expect `go` to be installed on runners - at least it is (meant to be) on github-hosted and grafana labs-hosted runners, and all `setup-argo` does is call `go env` to get some of the variables. although we did have an issue recently where it went missing heh.

also, if this is required (again, not sure it is) then the fix would be to have `setup-argo` itself call `setup-go` 🙂

mem:
> what's the problem that you saw? we'd expect go to be installed on runners - at least it is (meant to be) on github-hosted and grafana labs-hosted runners, and all setup-argo does is call go env to get some of the variables. although we did have an issue recently where it went missing heh.

You don't want to me talk about this, I'll go off on a rant about people's expectations about what CI should or shouldn't do 😉 

For the specific issue here, I had a similar expectation: Go is already there, this shouldn't be a problem. The shared action, trigger-argo-workflow, does this:

```
2024-10-12T02:41:31.6165536Z ##[group]Run ./_shared-workflows-trigger-argo-workflow/actions/setup-argo
2024-10-12T02:41:31.6165902Z with:
2024-10-12T02:41:31.6166075Z   cache-prefix: argo
2024-10-12T02:41:31.6166366Z   version: 3.5.1
2024-10-12T02:41:31.6166555Z ##[endgroup]
2024-10-12T02:41:31.6193804Z ##[group]Run actions/cache@3624ceb22c1c5a301c8db4169662070a689d9ea8
2024-10-12T02:41:31.6194125Z with:
2024-10-12T02:41:31.6194370Z   path: /opt/actions-runner/_work/sm-alerts/sm-alerts/bin/argo
2024-10-12T02:41:31.6194685Z   key: argo-Linux-X64-3.5.1
2024-10-12T02:41:31.6194907Z   enableCrossOsArchive: false
2024-10-12T02:41:31.6195132Z   fail-on-cache-miss: false
2024-10-12T02:41:31.6195350Z   lookup-only: false
2024-10-12T02:41:31.6195534Z   save-always: false
2024-10-12T02:41:31.6195713Z ##[endgroup]
2024-10-12T02:41:31.8386196Z Cache not found for input keys: argo-Linux-X64-3.5.1
2024-10-12T02:41:31.8436013Z ##[group]Run echo "OS=$(go env GOOS)" >> "$GITHUB_ENV"
2024-10-12T02:41:31.8436384Z [36;1mecho "OS=$(go env GOOS)" >> "$GITHUB_ENV"[0m               <--------------------------------------
2024-10-12T02:41:31.8436724Z [36;1mecho "ARCH=$(go env GOARCH)" >> "$GITHUB_ENV"[0m
2024-10-12T02:41:31.8451400Z shell: /usr/bin/sh -e {0}
2024-10-12T02:41:31.8451651Z ##[endgroup]
2024-10-12T02:41:31.8477034Z /opt/actions-runner/_work/_temp/505dabba-dce4-4b60-a243-e887d170d08f.sh: 1: go: not found <-----------------
2024-10-12T02:41:31.8478851Z /opt/actions-runner/_work/_temp/505dabba-dce4-4b60-a243-e887d170d08f.sh: 2: go: not found
```

It does not stop after that error, so the errors after that become nonsensical.

The specific job that is failing looks like this:

```
  release:
    name: release
    needs: [publish_images]
    if: ${{ always() && needs.publish_images.result == 'success' }}
    runs-on: ubuntu-amd64
    steps:
      # Remove this once trigger-argo-workflows is fixed.
      - name: Setup Go
        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
        with:
          go-version: '^1.23.1'

      - name: Trigger argo workflow
        uses: grafana/shared-workflows/actions/trigger-argo-workflow@f0dd3480fa3e657d741dd9e8d9b999cfb61fc713
        with:
          namespace: synthetic-monitoring-cd
          workflow_template: deploy-${{ steps.info.outputs.repository_name }}
          extra_args: "--generate-name deploy-${{ steps.info.outputs.repository_name }}-"
          parameters: |
            mode=${{ (inputs.mode == 'prod' && 'release') || (inputs.mode == 'dev' && 'dev') }}
            image=${{ needs.publish_images.outputs.image_name }}:${{ needs.publish_images.outputs.image_version }}
            commit=${{ github.sha }}
            commit_author=${{ github.actor }}
            commit_link=${{ github.event.repo.url }}/commit/${{ github.sha }}
```

That's basically the entire job: trigger an argo workflow.

Without the first step, it fails in the way shown above.

The effect of the first step is to have `setup-go` run before `setup-argo` (this change).

Why it's failing to find the go binary that is available in the runner, 🤷🏽 

Since `setup-argo` _does neet_ Go, and since the shared action _is_ using `setup-go` to get that, I figured this is a correct fix regardless of that other question.

guicaulada:
> Why it's failing to find the go binary that is available in the runner, 🤷🏽

The problem is caused by a mismatch between amd64 and x64, since our runners have to support both arm64 and amd64 the installation scripts were modified to support both, however some tools use amd64 others x64, same goes for arm64 and aarch64, so during image creation the variable `GOROOT_1_23_X64` was wrongly pointing to a `.../amd64` installation path, but the tool was installed on `.../x64`.

We have updated the script so it retrieves the architecture used in the [version manifest](https://github.com/actions/go-versions/blob/main/versions-manifest.json) and use it consistently through the instalation and setup of environment variables. Installation already used the archicteture from the versions manifest but environment setup didn't, it was hardcoded. Since the upstream actions/runner-images repo doesn't support multiple architectures this wasn't a problem for them but it is for us.

mem:
> > Why it's failing to find the go binary that is available in the runner, 🤷🏽
> 
> The problem is caused by a mismatch between amd64 and x64, [...]

Thank you for the explanation!

Just for my understanding: are you saying this PR is no longer necessary? If the answer is yes, I will be happy to open another one to remove [the following lines from the shared action](https://github.com/grafana/shared-workflows/blob/main/actions/trigger-argo-workflow/action.yaml#L59-L65):

```
    - name: Setup go
      uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
      with:
	@@ -64,6 +61,9 @@ runs:
          actions/trigger-argo-workflow/go.sum
        go-version-file: "_shared-workflows-trigger-argo-workflow/actions/trigger-argo-workflow/go.mod"
```

mem:
> I believe the workaround is still needed for this action, if it is to be used in amd64 runners, the PR is still necessary, at least until the new images are available in production.

It's necessary but you are not approving it (this PR)... do I need to do something else to get that to happen?

> I wouldn't revert this change after the new images are there though, if you're using Go in the action I would call setup-go beforehand to ensure you're using the correct version instead of the default version in the runner, which can change.

If I set up my workflow as:

```
setup-go
trigger-argo-workflow
```

that effectively translates to:

```
setup-go (with the version I specify)
...
setup-argo
setup-go (with the version specified in trigger-argo-workflow)
...
```

I have no idea if setup-go is going to do this twice? or if it's going to stop because Go is already there? or if it's going to upgrade the Go version (if the second one is newer than the first one)? or if it's going to simply overwrite the first one?

The docs for setup-go don't mention this (or I missed it).

PR #464, opened 2024-10-14
feat: add github oidc jwt token as a header

eloymg:
Added Github OIDC JWT Token to Proxy-Authorization-Token header

### DIFF
--- a/actions/get-vault-secrets/action.yaml
+++ b/actions/get-vault-secrets/action.yaml
@@ -61,7 +61,13 @@ runs:
         REPO_SECRETS: ${{ inputs.repo_secrets }}
         COMMON_SECRETS: ${{ inputs.common_secrets }}
         REPO: ${{ github.repository }}
-
+    - id: get-github-jwt-token
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const jwt = await core.getIDToken();
+          core.setSecret(jwt);
+          core.setOutput("github_jwt",jwt);
       # Get the secrets
     - name: Import Secrets
       id: import-secrets
@@ -74,5 +80,6 @@ runs:
         jwtGithubAudience: "https://vault-github-actions.grafana-${{ inputs.vault_instance }}.net"
         extraHeaders: |
           Proxy-Authorization: Bearer ${{ steps.vault-iap-auth.outputs.id_token }}
+          Proxy-Authorization-Token: Bearer ${{ steps.get-github-jwt-token.outputs.github_jwt }}
         secrets: |
           ${{ steps.translate-secrets.outputs.secrets }}
--- a/actions/get-vault-secrets/action.yaml
+++ b/actions/get-vault-secrets/action.yaml
@@ -63,7 +63,7 @@ runs:
         role: vault-github-actions
         path: github-actions-oidc
         method: jwt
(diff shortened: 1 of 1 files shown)


actions/get-vault-secrets/action.yaml
```diff
@@ -62,6 +62,14 @@ runs:
         COMMON_SECRETS: ${{ inputs.common_secrets }}
         REPO: ${{ github.repository }}
 
+    - id: get-github-jwt-token
+      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+      with:
+        script: |
+          const jwt = await core.getIDToken();
+          core.setSecret(jwt);
+          core.setOutput("github_jwt",jwt);
```
iainlane, line 71: super nit (and not required, I just think it looks nicer this way 😁)

```suggestion
          core.setOutput("github-jwt",jwt);
```

actions/get-vault-secrets/action.yaml
```diff
@@ -74,5 +82,6 @@ runs:
         jwtGithubAudience: "https://vault-github-actions.grafana-${{ inputs.vault_instance }}.net"
         extraHeaders: |
           Proxy-Authorization: Bearer ${{ steps.vault-iap-auth.outputs.id_token }}
+          Proxy-Authorization-Token: Bearer ${{ steps.get-github-jwt-token.outputs.github_jwt }}
```
iainlane, line 85:
```suggestion
          Proxy-Authorization-Token: Bearer ${{ steps.get-github-jwt-token.outputs.github-jwt }}
```

actions/get-vault-secrets/action.yaml
```diff
@@ -62,6 +62,14 @@ runs:
         COMMON_SECRETS: ${{ inputs.common_secrets }}
         REPO: ${{ github.repository }}
 
+    - id: get-github-jwt-token
+      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+      with:
+        script: |
+          const jwt = await core.getIDToken();
```
iainlane, line 69: let's set an audience though, so that on the other side we can validate the audience. it prevents leaked tokens from other workflows being given to this endpoint.

```suggestion
          const jwt = await core.getIDToken("vault-github-actions-grafana-${{ inputs.instance }}");
```

(the given audience is an example - you can change it if you want 🙂)

actions/get-vault-secrets/action.yaml
```diff
@@ -62,6 +62,14 @@ runs:
         COMMON_SECRETS: ${{ inputs.common_secrets }}
         REPO: ${{ github.repository }}
 
+    - id: get-github-jwt-token
+      uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+      with:
+        script: |
+          const jwt = await core.getIDToken("vault-github-actions-grafana");
```
iainlane, line 69: (not required from my perspective) - I would put the instance in here. It just means you can't use a `dev` ID token with the `ops` proxy and the other way around. That's what "audience" means to me: who is this token for? And we can be quite specific in this case 🙂

iainlane (approved):
👍

iainlane (commented):
I'm seeing unrelated changes now so removing my approval.

Suggest doing tests in a separate branch :-)

PR #471, opened 2024-10-16
feat: added github oidc token as a header

eloymg:
added github oidc token as a header. This allow us to remove IAP proxy from the auth process beacouse we validate the github OIDC token directly.


iainlane (approved):

iainlane:
I'd expand the commit message just _slightly_ to say that this allows us to remove IAP from the auth process by having a separate proxy.

PR #472, opened 2024-10-16
fix(ci): only run get-vault-secrets ci from `grafana` PRs

iainlane:
The `get-vault-secrets` action only works when run from a `grafana` repository, so skip this test if the PR is from a different repository. We will still get a run of this workflow for the change before merging, as we use merge queues. We can see an example of this failing in https://github.com/grafana/shared-workflows/pull/147.

I've created this PR itself from a fork, to test that this works. I pushed the same change to https://github.com/grafana/shared-workflows/pull/473 to see that we still ran the test from a non-fork. 👍

### DIFF
--- a/.github/workflows/test-get-vault-secrets.yaml
+++ b/.github/workflows/test-get-vault-secrets.yaml
@@ -12,6 +12,7 @@ on:
     paths:
       - "actions/get-vault-secrets/**"
       - ".github/workflows/test-get-vault-secrets.yaml"
+
   merge_group:
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-get-vault-secrets.yaml
```diff
@@ -28,6 +29,12 @@ jobs:
           - invalid
     runs-on: ubuntu-latest
 
+    # The `get-vault-secrets` action only works when run from a `grafana`
+    # repository, so skip this test if the PR is from a different repository. We
+    # will still get a run of this workflow for the change before merging, as we
+    # use merge queues.
```
iainlane, line 35: I'm assuming here that a merge queue branch will be in the target repository (and so `get-vault-secrets` will work) but I don't actually know this. We'll find out when we try, if the CI still runs and passes 🙃.

zerok (approved):

PR #479, opened 2024-10-16
chore(ci): run `pull_request` workflows when undrafting

iainlane:
We have the GitHub Actions bot user create release PRs, and this user doesn't trigger workflows. Fortunately, since we create the PRs in draft mode and a human undrafts to make a release, we can use this as a trigger to run workflows. This only works if the workflows trigger when undrafted. Add that here. This is the same approach we're taking in `wait-for-github` and it works fine there.


eloymg (approved):
LGTM

PR #488, opened 2024-10-17
feat(check-drone-sig): init reusable workflow

zzehring:
This reusable workflow is intended to be used by public, trusted drone-enabled repos to ensure changes to drone config are properly signed when the config it updated. Without this check, users can commit changes to the config without resigning the config file and cause builds to require manual approval to run.

### DIFF
--- a/.github/workflows/check-drone-signature.yaml
+++ b/.github/workflows/check-drone-signature.yaml
@@ -1,0 +1,54 @@
+name: Drone CI Signature Check
+
+on:
+  workflow_call:
+    inputs:
+      drone_config_path:
+        description: 'Path to the Drone CI configuration file'
+        required: true
+        type: string
+        default: '.drone.yml'
+      drone_server:
+        description: 'Drone CI server URL'
+        required: true
+        type: string
+        default: 'https://drone.grafana.net'
+
+jobs:
+  check-drone-signature:
+    permissions:
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+
+      - name: Set up Drone CLI
+        run: |
+          curl -L https://github.com/drone/drone-cli/releases/latest/download/drone_linux_amd64.tar.gz | tar zx
+          sudo install -t /usr/local/bin drone
+
+      - name: Retrieve drone-signature-checker token
+        uses: grafana/shared-workflows/actions/get-vault-secrets@main
+        with:
+          common_secrets: |
+            DRONE_TOKEN=drone-sign-check-machine-user:token
+
+      - name: Check Drone signature
+        run: |
+          # Run drone sign command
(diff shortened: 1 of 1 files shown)


zzehring:
Tested here: https://github.com/grafana/synthetic-monitoring-agent/actions/runs/11393520767/job/31701958563

.github/workflows/check-drone-signature.yaml
```diff
@@ -0,0 +1,52 @@
...
+          # Run drone sign command
+          drone sign --save ${{ github.repository }} ${{ inputs.drone_config_path }}
+
+          # Check if there are any changes to the Drone config file
+          if git diff --quiet ${{ inputs.drone_config_path }}; then
+            echo "Drone CI configuration signature is valid."
+          else
+            echo "Drone CI configuration signature was updated."
+            echo "Please sign the updated .drone.yml file and commit."
+            git --no-pager diff ${{ inputs.drone_config_path }}
```
zerok, line 47: I think showing that diff might encourage people to just copy&paste the output into their configs instead of actually running `drone sign` locally. WDYT?

.github/workflows/check-drone-signature.yaml
```diff
@@ -0,0 +1,52 @@
...
+      contents: read
+      id-token: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+
+      - name: Set up Drone CLI
+        run: |
+          curl -L https://github.com/drone/drone-cli/releases/latest/download/drone_linux_amd64.tar.gz | tar zx
```
zerok, line 27: Should this perhaps be pinned to a specific version of the CLI?

zzehring: I'm not sure drone supports downloading a specific version. See [installation instructions docs](https://docs.drone.io/cli/install/).

.github/workflows/check-drone-signature.yaml
zzehring: Ahhh now that I think about it, we shouldn't surface that message/the hash. Since the point of the signature is to see if the drone config has changed and only "authorized" users would be able to sign and get that hash. 

External contributors also wouldn't be able to change the workflow to print the signature as it needs the drone token (from vault secret) to do so and getting secrets in forked pull-requests doesn't work.

I'll remove the print.

.github/workflows/check-drone-signature.yaml
Pokom: You should be able to do something to infer the version. IE I can download a specific version with a URL like this: https://github.com/harness/drone-cli/releases/download/v1.8.0/drone_linux_amd64.tar.gz

So you should be able to parameterize the version.

zerok (approved):
Thank you 😄

.github/workflows/check-drone-signature.yaml
zerok: Awesome, thank you 🙂

PR #497, opened 2024-10-23
feat: add validate-policy-bot-action

Elfo404:
WIP


zerok (approved):

iainlane:
I would move that out to a public repo (probably its own one?), that'd be great

iainlane:
One thing I just realised. Since we use merge queues in _this_ repository the policy bot isn't going to be very helpful. We should still provide a shared workflow for folks that don't use them, but I'm not sure about the generated one in this PR. 🤔

https://github.com/palantir/policy-bot/issues/800

Elfo404:
@iainlane  Removed the check for this repo and updated the README of the action. mind a sanity check?

PR #508, opened 2024-10-30
feat: allow skipping QEMU in Docker build&push action

mariomac:
Qemu slowdowns compilation by orders or magnitude.

At least in Go applications, buildx should work well without QEMU for multiarch builds.

This PR adds a configuration option to disable Qemu installation.


mariomac:
Closing PR. After trying in GitHub, it does not seem to provide any performance improvement.

PR #512, opened 2024-11-01
chore(renovate): update Renovate itself less frequently (weekly)

iainlane:
Renovate pushes out new images most days, sometimes multiple times a day. We don't need to track these so closely, so let's look for these particular updates only once a week - early on Monday mornings.


zerok (approved):
Thank you ❤️

Issue #482, opened 2024-10-17
trigger-argo-workflow fails to run because go is not available


mem:
using this job in a workflow:

```
  release:
    name: release
    runs-on: ubuntu-amd64
    steps:
      - name: Trigger argo workflow
        uses: grafana/shared-workflows/actions/trigger-argo-workflow@f0dd3480fa3e657d741dd9e8d9b999cfb61fc713
        with:
          namespace: ...
          workflow_template: ...
          extra_args: ...
          parameters: |
            ...
```

(removed the arguments because they are not relevant for the issue)

Fails with:

```
2024-10-12T02:41:31.6165536Z ##[group]Run ./_shared-workflows-trigger-argo-workflow/actions/setup-argo
2024-10-12T02:41:31.6165902Z with:
2024-10-12T02:41:31.6166075Z   cache-prefix: argo
2024-10-12T02:41:31.6166366Z   version: 3.5.1
2024-10-12T02:41:31.6166555Z ##[endgroup]
2024-10-12T02:41:31.6193804Z ##[group]Run actions/cache@3624ceb22c1c5a301c8db4169662070a689d9ea8
2024-10-12T02:41:31.6194125Z with:
2024-10-12T02:41:31.6194370Z   path: /opt/actions-runner/_work/sm-alerts/sm-alerts/bin/argo
2024-10-12T02:41:31.6194685Z   key: argo-Linux-X64-3.5.1
2024-10-12T02:41:31.6194907Z   enableCrossOsArchive: false
2024-10-12T02:41:31.6195132Z   fail-on-cache-miss: false
2024-10-12T02:41:31.6195350Z   lookup-only: false
2024-10-12T02:41:31.6195534Z   save-always: false
2024-10-12T02:41:31.6195713Z ##[endgroup]
2024-10-12T02:41:31.8386196Z Cache not found for input keys: argo-Linux-X64-3.5.1
2024-10-12T02:41:31.8436013Z ##[group]Run echo "OS=$(go env GOOS)" >> "$GITHUB_ENV"
2024-10-12T02:41:31.8436384Z [36;1mecho "OS=$(go env GOOS)" >> "$GITHUB_ENV"[0m               <--------------------------------------
2024-10-12T02:41:31.8436724Z [36;1mecho "ARCH=$(go env GOARCH)" >> "$GITHUB_ENV"[0m
2024-10-12T02:41:31.8451400Z shell: /usr/bin/sh -e {0}
[...]

PR #519, opened 2024-11-06
fix(setup-argo): do not require go binary

zerok:
Fixes https://github.com/grafana/shared-workflows/issues/482

### DIFF
--- a/actions/setup-argo/action.yaml
+++ b/actions/setup-argo/action.yaml
@@ -31,8 +31,24 @@ runs:
       if: steps.cache.outputs.cache-hit != 'true'
       shell: sh
       run: |
-        echo "OS=$(go env GOOS)" >> "$GITHUB_ENV"
-        echo "ARCH=$(go env GOARCH)" >> "$GITHUB_ENV"
+        OS=$(echo "${{ runner.os }}" | tr '[:upper:]' '[:lower:]')
+        ARCH=$(echo "${{ runner.arch }}" | tr '[:upper:]' '[:lower:]')
+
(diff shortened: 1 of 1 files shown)


zerok:
The problem is that `setup-argo` currently assumes that a previous setup has already run `setup-go`. IMO we have two options here: Do the mapping ourselves from the `runner.os/arch` stuff, or install Go inside the action for the single purpose of doing the mapping for us. 

IMO, doing away with this "artificial" dependency is the better approach as either will fail if argo decided to change their package names but one does not require Go to be installed 🙂

iainlane:
> assumes that a previous setup has already run `setup-go`.

Not quite - it assumes `go` is present on the runners, which is true for GitHub's images at least.

Have you got a concrete failure to point to?

zerok:
> > assumes that a previous setup has already run `setup-go`.
> 
> Not quite - it assumes `go` is present when the step is run (no matter how it got there), which is true for GitHub's images at least.
> 
> Have you got a concrete failure to point to?

https://github.com/grafana/enghub/actions/runs/11701429721/job/32595152894 and probably the one listed in the original ticket.

iainlane:
`go` is supposed to be installed on our internal runners but isn't due to a bug... but as far as I know the known blockers to us rolling out a fixed image are resolved now so we should be able to get back to a state where it is always there. @guicaulada is that right?

zerok:
Working around this for now until the images are fixed 🙂

iainlane:
FWIW the internal `arm64` runners should be fine

PR #530, opened 2024-11-12
feat: add action to log into Docker Hub pull-through cache

iainlane:
Grafana Labs has a pull-through cache hosted in GAR to avoid hititng Docker Hub for every image request, which is something that is quite subject to rate limiting.

This adds a new composite action to make logging into this registry easier.


iainlane:
The team didn't want to do this.

PR #546, opened 2024-11-25
feat(lint-pr-title): load configuration files from our directory

iainlane:
`commitlint` configs can have an `extends` key which refers to a module. There's some runtime support for resolving this module, which requires it to be discoverable either globally or from the current directory (e.g. via a `node_modules` contained in it). We're currently trying to load modules from anywhere on the filesystem, via a configuration flag, and failing when they're not inside the Action's own directory tree.

Here we're switching to the latter approach: copying the user-supplied config file to our directory. This means we can keep the code testable and not have action-specific installation code.

This is done by adding a new method of creating temporary files that are automatically cleaned up when they go out of scope, using [`AsyncDisposable`][e-r-m].

[e-r-m]: https://github.com/tc39/proposal-explicit-resource-management


zerok (approved):

iainlane:
I'm going to merge #564, then any updates it creates, and then rebase this, to get the dependency updates out of this PR. 👍

PR #548, opened 2024-11-27
chore: update readme when a new release is available

dsotirakis:
We need this feature so the README.md guide always contains the latest version of the action.

Currently our workflow READMEs are a bit misleading since we tell people to use `main`, which will cause backwards compatibility issues.

[Working example](https://github.com/grafana/shared-workflows/pull/571/files).

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,4 +153,4 @@ jobs:
 
 ````
 
-Every semver-like string nested in the above release-please block, will be updated with the new semver once it's released.
+Every semver-like string between the `x-release-please-start-version` and `x-release-please-end-version` comments will be updated with the new version whenever the component is released.
--- a/README.md
+++ b/README.md
@@ -108,3 +108,50 @@ More about how the upstream action works can be found [here](https://github.com/
 ### Add new components to Release Please config file
 
 In order for components to be released, they must be in the [release-please-config.json](./release-please-config.json) file. Always ensure new components are added to this file.
+
+#### Automatically update README files with the latest version
+
+As part of using the latest version of an action, it's a good practice to update the README file of each shared workflow every time there is a new release.
+
+To do so, after adding the name of the new action which needs to be released in the [release-please-config.json](./release-please-config.json) file, the `"extra-files"` field should also be added and include the `README.md` like:
+
+```json
+  ...
+  "packages": {
+    "actions/my-new-action": {
+      "package-name": "my-new-action",
+      "extra-files": ["README.md"]
+    },
+  }
+```
+
+Also, the following block should be added in the README file which will be responsible for updating the version with a new one:
+
+`README.md`:
+````
+# my-new-action
+
+This is my new action which does awesome stuff!
+
+<!-- x-release-please-start-version --> # beginning of the release please block
+
+```yaml
+name: My new action
+on:
+  pull_request:
+
+jobs:
+  my-new-action:
+    runs-on: ubuntu-latest
+
+    steps:
+      - id: do-stuff
+        uses: grafana/shared-workflows/actions/my-new-action@my-new-action-v1.0.0
+
+```
+
+<!-- x-release-please-end-version --> # end of the release please block
+
+````
+
+Every semver-like string nested in the above release-please block, will be updated with the new semver once it's released.
--- a/README.md
+++ b/README.md
@@ -114,7 +114,6 @@ Once the component is added, as part of using the latest version of an action, i
 To do so, after adding the name of the new action which needs to be released in the [release-please-config.json](./release-please-config.json) file, the `"extra-files"` field should also be added and include the `README.md` like:
 
 ```json
-  ...
   "packages": {
     "actions/my-new-action": {
       "package-name": "my-new-action",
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -108,3 +108,51 @@ More about how the upstream action works can be found [here](https://github.com/
 ### Add new components to Release Please config file
 
 In order for components to be released, they must be in the [release-please-config.json](./release-please-config.json) file. Always ensure new components are added to this file.
+
+#### Automatically update README files with the latest version
```
iainlane, line 112: I would merge these sections somehow, and tell people what to put in the config file. It's a bit weird to have "add to this config file" and then the next section "add some more to that same file" heh

iainlane (commented):
Cool, looks like a good change 👍

Can we have a linter which ensures this file stays correct? Otherwise people (like me) will forget. 😢

dsotirakis:
> Can we have a linter which ensures this file stays correct? Otherwise people (like me) will forget. 😢

Yeah, I want to create a new PR with the remaining actions and then create that lint rule if that's okay?

iainlane:
> > Can we have a linter which ensures this file stays correct? Otherwise people (like me) will forget. 😢
> 
> Yeah, I want to create a new PR with the remaining actions and then create that lint rule if that's okay?

OK if you want to do it that way, that's fine

README.md
dsotirakis: I thought that having this as a header 3 could mean that it's optional in case someone doesn't want to do it but I think what you say is fair, I'll merge them together.

README.md
```diff
@@ -108,3 +108,49 @@ More about how the upstream action works can be found [here](https://github.com/
...
+      "extra-files": ["README.md"]
+    },
+  }
+```
+
+Also, the following block should be added in the README file which will be responsible for updating the version with a new one:
+
+`README.md`:
+
+````
```
iainlane, line 130:
```suggestion
````markdown
```

(think this works)

README.md
```diff
@@ -108,3 +108,49 @@ More about how the upstream action works can be found [here](https://github.com/
 ### Add new components to Release Please config file
 
 In order for components to be released, they must be in the [release-please-config.json](./release-please-config.json) file. Always ensure new components are added to this file.
+
+Once the component is added, as part of using the latest version of an action, it's a good practice to update the README file of each shared workflow every time there is a new release.
```
iainlane, line 112: Along the lines of consistency, I would be a bit more direct and say that it is needed

```suggestion

`README` files for each component should have embedded versions updated every time there is a new release.
```

README.md
```diff
@@ -108,3 +108,49 @@ More about how the upstream action works can be found [here](https://github.com/
...
+
+To do so, after adding the name of the new action which needs to be released in the [release-please-config.json](./release-please-config.json) file, the `"extra-files"` field should also be added and include the `README.md` like:
```
iainlane, line 114: This could be made a lot shorter and still be fine I think

```suggestion
Add a new entry which looks like this:
```

README.md
```diff
@@ -108,3 +108,49 @@ More about how the upstream action works can be found [here](https://github.com/
...
+Also, the following block should be added in the README file which will be responsible for updating the version with a new one:
+
+`README.md`:
+
+````
+# my-new-action
+
+This is my new action which does awesome stuff!
+
+<!-- x-release-please-start-version --> # beginning of the release please block
```
iainlane, line 135: If people copy and paste this with the `# beginning of the release please block` text, that will show up in their README. Might suggest the following 3...

```suggestion
<!-- x-release-please-start-version -->
```

README.md
```diff
@@ -108,3 +108,49 @@ More about how the upstream action works can be found [here](https://github.com/
...
+  my-new-action:
+    runs-on: ubuntu-latest
+
+    steps:
+      - id: do-stuff
+        uses: grafana/shared-workflows/actions/my-new-action@my-new-action-v1.0.0
+
+```
+
+<!-- x-release-please-end-version --> # end of the release please block
```
iainlane, line 152:
```suggestion
<!-- x-release-please-end-version -->
```

README.md
```diff
@@ -108,3 +108,49 @@ More about how the upstream action works can be found [here](https://github.com/
...
+
+````
+
+Every semver-like string nested in the above release-please block, will be updated with the new semver once it's released.
```
iainlane, line 156:
```suggestion
Every semver-like string between the `x-release-please-start-version` and `x-release-please-end-version` comments will be updated with the new version whenever the component is released.
```

release-please-config.json
```diff
@@ -54,49 +54,40 @@
...
-      "package-name": "push-to-gcs"
-    },
-    "actions/send-slack-message": {
-      "package-name": "send-slack-message"
-    },
-    "actions/setup-argo": {
-      "package-name": "setup-argo"
-    },
-    "actions/setup-conftest": {
-      "package-name": "setup-conftest"
```
iainlane, line 99: what's happened to these?

dsotirakis: wat, I don't know how this happened. will add them back, thanks for spotting

README.md
dsotirakis: wanted to make sure that people know what the change is really. it's clear from the docs I guess, so fine to remove it.

PR #568, opened 2024-11-27
chore(ci): pin dependencies too

iainlane:
We're using the default behaviour of pinning only dev dependencies. The effect of this can be seen in 49a4348fb372ec3a866f275ce7a98fb21c2788c5.

Renovate has [an essay][dependency-pinning] discussing this whole area. The [summary] is that for our case, developing an application not a library, it would be more robust to [pin runtime dependencies][pinDependencies] too.

[dependency-pinning]: https://docs.renovatebot.com/dependency-pinning/
[pinDependencies]: https://docs.renovatebot.com/presets-default/#pindependencies
[summary]: https://docs.renovatebot.com/dependency-pinning/#so-whats-best

### DIFF
--- a/.github/renovate-config.json5
+++ b/.github/renovate-config.json5
@@ -16,7 +16,7 @@
   dependencyDashboard: false,
   enabledManagers: ["bun", "custom.regex", "github-actions", "npm"],
   forkProcessing: "enabled",
-  globalExtends: ["config:best-practices"],
+  globalExtends: [":pinDependencies", "config:best-practices"],
   onboarding: false,
   osvVulnerabilityAlerts: true,
   packageRules: [


.github/renovate-config.json5
```diff
@@ -16,7 +16,7 @@
   dependencyDashboard: false,
   enabledManagers: ["bun", "custom.regex", "github-actions", "npm"],
   forkProcessing: "enabled",
-  globalExtends: ["config:best-practices"],
+  globalExtends: [":pinDependencies", "config:best-practices"],
```
dsotirakis, line 19: it's not clear from the docs if we really need the : in the front? 🤔

iainlane: [It's part of the name](https://docs.renovatebot.com/presets-default/#pindependencies). What's making you think otherwise?

Check the [logs for the run on this PR](https://github.com/grafana/shared-workflows/actions/runs/12051300936/job/33602021745?pr=568) too - it tells you what it would have done.

dsotirakis: yeah couldn't understand if we should have it there or not, but since it's working I guess that's the right way to go 😁 thought it was a c/p issue!

dsotirakis (approved):

.github/renovate-config.json5
iainlane: heh. thanks for being diligent!

PR #573, opened 2024-11-27
feat(dependabot-gar-login): create composite action

guicaulada:
The dependabot job for updating Docker images in our private registries has been failing.

From the job logs we can see dependabot fails on authenticating to two different repositories that we use as sources.

These failures also seem to make the successfully-queried images not create PRs even though the logs indicate they do.

ref: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/configuring-access-to-private-registries-for-dependabot#storing-credentials-for-dependabot-to-use

This PR suggests creating a shared action that would allow multiple repositories to solve this issue easily by creating a workflow that updates a token for Dependabot safely using OIDC.

### DIFF
--- a/actions/dependabot-gar-login/README.md
+++ b/actions/dependabot-gar-login/README.md
@@ -1,0 +1,39 @@
+# Dependabot GAR Login
+
+Updates the token for Google Artifact Registry and sets it as a secret for Dependabot at the repository level.
+
+The token can be refreshed and set to the `DEPENDABOT_GAR_TOKEN` secret every 50 minutes to ensure Dependabot can access Google Artifact Registry without interruptions.
+
+## Example
+
+```yaml
+name: Update Google Artifact Registry Token
+
+on:
+  schedule:
+    # Update every 50 minutes
+    - cron: "*/50 * * * *"
+  workflow_dispatch:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Update Artifact Registry Token
+        uses: ./.github/actions/update-artifact-registry-token
+```
+
+## Configuring Dependabot
+
+To use the secret in your dependabot.yml, add the following configuration:
+
+```yaml
+registries:
+  google-artifact-registry:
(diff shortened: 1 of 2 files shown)


actions/dependabot-gar-login/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Update Artifact Registry Token
+        uses: ./.github/actions/update-artifact-registry-token
```
zerok, line 26: Please already point to the final destination of this action.

actions/dependabot-gar-login/README.md
```diff
@@ -0,0 +1,39 @@
...
+    # Update every 50 minutes
+    - cron: "*/50 * * * *"
+  workflow_dispatch:
+
+jobs:
+  update:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
```
zerok, line 23: Do you need content-read if the action does not come from a relative path? I'm also wondering, don't you need some kind of write-permissions to actually modify a secret?

zerok: Ah, makes sense. Thank you :)

actions/dependabot-gar-login/action.yaml
```diff
@@ -0,0 +1,44 @@
+name: "Update Google Artifact Registry Token"
+description: "Updates the token for Google Artifact Registry and sets it as a secret for Dependabot at the repository level."
+
+runs:
+  using: "composite"
+  steps:
+    - id: get-secrets
+      uses: grafana/shared-workflows/actions/get-vault-secrets@main
```
iainlane, line 8: This needs to refer to the action locally. See the top-level readme for how to do this

guicaulada:
I decided to close this in favor of having it on the private repository that needs it, currently I couldn't find the need for public repos to have this, having it in the private repo it's being used on will make development and maintenance easier.

If we have the need for this in multiple repositories in the future we can revisit and move it to shared-workflows.

Issue #478, opened 2024-10-16
PRs can still be merged with failing workflows


iainlane:
We had [this failing run](https://github.com/grafana/shared-workflows/actions/runs/11369868620) but we still merged #147 and it broke users.

It's because this check isn't required. Conditional checks can't simply be made required, or else they will block pull requests which don't change the files they trigger on.

Possible fixes:

1. Make checks unconditional, just run them always
2. Move the conditionality to runtime using a "changed files" type of action
3. Use our deployment of https://github.com/palantir/policy-bot and make these checks conditional _if triggered_ by writing a `.policy.yml` containing rules that replicate the file triggers, or using a generator (we have one internally in `deployment_tools` - we could open source that by moving it to this repo).

Elfo404:
Option 3 should not be viable due to https://github.com/grafana/shared-workflows/pull/497#issuecomment-2455223956

dblinkhorn:
What about if we implement an "orchestrator" workflow that analyzes the changed files and decides which test workflows need to be run? This workflow can be set as a required check, but if none of the tests are triggered it will just return success.

```yaml
name: Test Workflow Orchestrator

on:
  push:
    branches: [main]
  pull_request:

permissions:
  contents: read
  pull-requests: read 

jobs:
# check the changed files and set outputs used to trigger individual test jobs below
  changes:
    runs-on: ubuntu-latest
    outputs:
      vault: ${{ steps.filter.outputs.vault }}
      gar: ${{ steps.filter.outputs.gar }}
      # etc...
    steps:
      - id: filter
        uses: dorny/paths-filter@de90cc6d5d3da4d2a4f21f4e0f6c1e73948b9a6   # pin v3.0.2
        with:
          filters: |
            vault:
              - 'actions/get-vault-secrets/**'
              - '.github/workflows/test-get-vault-secrets.yaml'
            gar:
              - 'actions/login-to-gar/**'
              - '.github/workflows/test-login-to-gar.yaml'
            # etc...

# invoke the tests which are needed based on changed files in diff

  vault-test:
    needs: changes
    if: needs.changes.outputs.vault == 'true'
    uses: ./.github/workflows/test-get-vault-secrets.yaml
    permissions:           # forward permissions to called workflow if needed
      contents: read
      id-token: write
    secrets: 

  gar-test:
    needs: changes
    if: needs.changes.outputs.gar == 'true'
    uses: ./.github/workflows/test-login-to-gar.yaml

# and so on...
```

[...]

PR #593, opened 2024-11-29
fix: make check-for-non-releasable-actions action run on any actions/ change

dsotirakis:
After https://github.com/grafana/shared-workflows/pull/588 got merged, I realized that the action wasn't triggered in the automated release-please PRs, even if the CHANGELOG changes, which lives in `actions/` directory.

This PR should fix the issue, by adding the double-start glob pattern.

### DIFF
--- a/.github/workflows/check-for-non-releasable-actions.yaml
+++ b/.github/workflows/check-for-non-releasable-actions.yaml
@@ -2,7 +2,7 @@ name: Check for non-releasable actions
 on:
   pull_request:
     paths:
-      - actions/
+      - actions/**
       - .github/workflows/check-for-non-releasable-actions.yaml
       - release-please-config.json
     types:
@@ -15,7 +15,7 @@ on:
     branches:
       - main
     paths:
-      - actions/
+      - actions/**
       - .github/workflows/check-for-non-releasable-actions.yaml
       - release-please-config.json
 


iainlane (commented):
Wasn't the point to trigger only if an action is added or removed? I [made a comment asking if that was how it worked](https://github.com/grafana/shared-workflows/pull/588/files#r1863275465), and it made sense to me that if it did then this was a good way to behave. The file should only be able to drift when the action changes, it's modified itself or when actions are added/removed, right?

Said in a different way, I'm not sure that it _should_ have run when #588 got merged.

Or doesn't it work like that after all?

dsotirakis:
> Wasn't the point to trigger only if an action is added or removed? I [made a comment asking if that was how it worked](https://github.com/grafana/shared-workflows/pull/588/files#r1863275465), and it made sense to me that if it did then this was a good way to behave. The file should only be able to drift when the action changes, it's modified itself or when actions are added/removed, right?
> 
> Said in a different way, I'm not sure that it _should_ have run when #588 got merged.
> 
> Or doesn't it work like that after all?

Yeah sorry that's right, I wasn't clear. The expectation should be that it should run only for additions/removals of actions in the `actions/` directory. It just occurred to me though that we should have this check in case someone renames the action? 🤔

iainlane:
> > Wasn't the point to trigger only if an action is added or removed? I [made a comment asking if that was how it worked](https://github.com/grafana/shared-workflows/pull/588/files#r1863275465), and it made sense to me that if it did then this was a good way to behave. The file should only be able to drift when the action changes, it's modified itself or when actions are added/removed, right?
> > Said in a different way, I'm not sure that it _should_ have run when #588 got merged.
> > Or doesn't it work like that after all?
> 
> Yeah sorry that's right, I wasn't clear. The expectation should be that it should run only for additions/removals of actions in the `actions/` directory. It just occurred to me though that we should have this check in case someone renames the action? 🤔

Yeah, be good to check how this existing check behaves! I wasn't totally sure but you gave me confidence 😄

dsotirakis:
Next test, change the target branch in [this PR](https://github.com/grafana/shared-workflows/pull/592/), to `fix-trigger-paths` (current PR) and it correctly doesn't trigger the action since nothing is touched.

dsotirakis:
So current implementation doesn't work, also single star glob pattern doesn't work either on renames or additions (tried both `actions/*` and `actions/*/`), so I guess double star should be fine. On any change that happens in `actions/`, we'll be triggering this action.

iainlane (approved):
Looks good. I'd consider removing the path filters completely, and then we can make the check a required one.

dsotirakis:
> Looks good. I'd consider removing the path filters completely

Hmm, I think paths could still be valuable for example to avoid running the check if we add new workflows (not actions) or touching stuff outside of the `actions/` directory.

> we can make the check a required one.

That's a good point though. Let's merge it now and revisit. I'll create an issue about it. Maybe we can also use the policy bot?

iainlane:
> Maybe we can also use the policy bot?

No afraid not, not since we're using the merge queue - https://github.com/grafana/shared-workflows/issues/478 is relevant to this. Running checks more often than needed is the best way we have right now sadly 😢

dsotirakis:
Ugh okay - then it makes sense to close this PR and open a new one which removes the paths completely!

dsotirakis:
closing in favor of: https://github.com/grafana/shared-workflows/pull/600

PR #608, opened 2024-12-02
feat(check-for-non-releasable-actions): find reusable workflows too

iainlane:
We need to release reusable workflows just like we do for Actions. The current linter didn't check for that, so it's added here. This required an extra library to parse the YAML, and GitHub scripts can't install new dependencies, so it's broken out to a small `internal/` composite action which can.

Now we have two Typescript actions, it made sense to centralise some of the setup. We have a shared `eslint` config now, and `bun` workspaces for each of the modules so they can cross-reference.


iainlane:
To be able to finish this PR, I need to figure out how to get `release-please` to release a reusable workflow. The problem is it wants to handle a directory (create a changelog for example, and update it when there are any changes in there) but these are just single files in `.github/workflows`.

PR #618, opened 2024-12-03
feat: introduce setup-crossplane action

Duologic:
Based on https://github.com/crossplane-contrib/xpkg-action which looks poorly maintained.

I'm using this action in https://github.com/grafana/grafana-crossplane-libsonnet to push Configuration packages to OCI repos (GHCR, Upbound Marketplace).


Duologic:
I've created a repo upstream: https://github.com/crossplane-contrib/setup-crossplane-action

PR #619, opened 2024-12-03
feat: introduce setup-jrsonnet action

Duologic:
I'm using a version of this action in https://github.com/grafana/grafana-crossplane-libsonnet but think it would be good to maintain this as a shared workflow.

### DIFF
--- a/actions/setup-jrsonnet/README.md
+++ b/actions/setup-jrsonnet/README.md
@@ -1,0 +1,16 @@
+# Setup Jrsonnet
+
+Setup Jrsonnet cli and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
+
+## Example
+
+<!-- x-release-please-start-version -->
+
+```
+uses: grafana/shared-workflows/actions/setup-jrsonnet@setup-jrsonnet-v1.0.0
+with:
+  version: 0.5.0-pre96-test # Version of the jrsonnet CLI to install.
+
+```
+
+<!-- x-release-please-end-version -->
--- a/actions/setup-jrsonnet/README.md
+++ b/actions/setup-jrsonnet/README.md
@@ -1,6 +1,6 @@
-# Setup Jrsonnet
+# Setup jrsonnet
 
-Setup Jrsonnet cli and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
+Setup jrsonnet CLI and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
 
 ## Example
 
--- a/actions/setup-jrsonnet/action.yaml
+++ b/actions/setup-jrsonnet/action.yaml
@@ -1,0 +1,64 @@
+name: Setup Jrsonnet
+description: Setup Jrsonnet cli and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
+
+inputs:
+  cache-prefix:
+    description: Prefix for the cache key.
+    default: jrsonnet
+
+  version:
+    description: |
+      Version of the jrsonnet CLI to install.
+    default: 0.5.0-pre96-test
+
+outputs:
+  cache-hit:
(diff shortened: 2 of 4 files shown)


zerok:
Would this perhaps even make sense as a separate repository? I imagine for instance that other Tanka users might be an audience here but not of the rest of the shared workflows.

Duologic:
How do we decide what should or shouldn't go into this repo?

zerok:
We have a broad scope defined here: https://github.com/grafana/shared-workflows/blob/main/CONTRIBUTING.md#scope-of-shared-workflows

I guess it comes down to what other repos would need this.

iainlane:
These `setup-` actions would be nice contributions to the upstream projects themselves too.

It doesn't necessarily need to block us adding them in here while that process is ongoing, though - that's one for folks' best judgement. 🙂

(Just making a comment/suggestion, not weighing in on the review - please carry on reviewing/discussing.)

actions/setup-jrsonnet/README.md
```diff
@@ -0,0 +1,16 @@
+# Setup Jrsonnet
+
+Setup Jrsonnet cli and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
```
zerok, line 3:
```suggestion
# Setup jrsonnet

Setup jrsonnet CLI and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
```

actions/setup-jrsonnet/action.yaml
```diff
@@ -0,0 +1,64 @@
+name: Setup Jrsonnet
+description: Setup Jrsonnet cli and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
```
zerok, line 2:
```suggestion
name: Setup jrsonnet
description: Setup jrsonnet CLI and add it to the PATH, this action will pull the binary from GitHub releases and store it in cache for the next run.
```

actions/setup-jrsonnet/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+        key: ${{ inputs.cache-prefix }}-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
+
+    - name: Map OS and ARCH to Jrsonnet release artifact
+      if: steps.cache.outputs.cache-hit != 'true'
+      shell: sh
+      run: |
+        echo "OS=$(go env GOOS)" >> "$GITHUB_ENV"
+        echo "ARCH=$(go env GOARCH)" >> "$GITHUB_ENV"
+
+    - name: Fetch Github Release Asset
```
zerok, line 37:
```suggestion
    - name: Fetch GitHub Release Asset
```

actions/setup-jrsonnet/action.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+  steps:
+    - name: Setup cache
+      id: cache
+      uses: actions/cache@6849a6489940f00c2f30c0fb92c6274307ccb58a # v4.1.2
+      with:
+        path: ${{ github.workspace }}/bin/jrsonnet
+        key: ${{ inputs.cache-prefix }}-${{ runner.os }}-${{ runner.arch }}-${{ inputs.version }}
+
+    - name: Map OS and ARCH to Jrsonnet release artifact
```
zerok, line 30:
```suggestion
    - name: Map OS and ARCH to jrsonnet release artifact
```

zerok (approved):
LGTM 🙂 Just some minor suggestions for consistency.

zerok:
@Duologic What do you think about, as Iain suggested, making this action available as part of the upstream repository? https://github.com/CertainLach/jrsonnet 🙂

Duologic:
I'm not opposed but like that we are currently in control of it, handing it over would require more back and forth in case we need to change it. Is it okay to wait with this for like a month so that get some mileage on this?

zerok:
Absolutely 🙂

Issue #69, opened 2024-03-21
`trigger-argo-workflow` resubmits the workflow multiple times after the workflow failure when `--wait` flag is used


fornfrey:
When submitting a workflow with `--wait` flag, the process blocks until the workflow is completed. However, when the submitted workflow fails the action treats it as a transient error and attempts to resubmit it. This causes multiple `AlreadyExists` errors to be printed to the job output.

This does not impact the workflow submission or execution (the workflow failed anyway) but makes the debugging process confusing. Sample output of a failed job:
```
Notice: running command
  time: "2024-03-21T09:49:12.896Z"
  level: INFO
  extraArgs:
  - --wait
Error: failed to run command, retrying
  time: "2024-03-21T10:00:05.717Z"
  level: ERROR
  error: exit status 1
  retry_in: 559.753636ms
time="2024-03-21T10:00:06.993Z" level=fatal msg="Failed to submit workflow: rpc error: code = AlreadyExists desc = workflows.argoproj.io \"my-workflow-1\" already exists, the server was not able to generate a unique name for the object"
Error: failed to run command, retrying
  time: "2024-03-21T10:00:06.995Z"
  level: ERROR
  error: exit status 1
  retry_in: 1.077149047s
time="2024-03-21T10:00:08.773Z" level=fatal msg="Failed to submit workflow: rpc error: code = AlreadyExists desc = workflows.argoproj.io \"my-workflow-1\" already exists, the server was not able to generate a unique name for the object"
```

dblinkhorn:
newbie question...is the workflow uri returned by the argo CLI prior to completion when using the `--wait` flag?

fornfrey:
@dblinkhorn, no, it does not. There is an issue for that too: https://github.com/grafana/shared-workflows/issues/70 
Thanks for addressing this one btw 🙇

dblinkhorn:
@fornfrey oops i posted that comment in the wrong issue! my bad. it was meant for #70. will repost!

PR #631, opened 2024-12-03
fix(trigger-argo-workflow): prevent unnecessary retries if a permanent error is encountered

dblinkhorn:
This fix utilizes the built-in [Permanent](https://pkg.go.dev/github.com/cenkalti/backoff/v4@v4.3.0#Permanent) function to abort the retry logic within `Run()` if a permanent error is returned by `runCmd()`.

Fixes [this](https://github.com/grafana/shared-workflows/issues/69) issue.

### DIFF
--- a/actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
+++ b/actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
@@ -33,8 +33,8 @@ type App struct {
 }
 
 var instanceToHost = map[string]string{
-	"dev":     "argo-workflows-dev.grafana.net:443",
-	"ops":     "argo-workflows.grafana.net:443",
+	"dev": "argo-workflows-dev.grafana.net:443",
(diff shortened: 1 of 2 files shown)


actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
```diff
@@ -136,6 +153,10 @@ func (a *App) Run(md GitHubActionsMetadata) error {
 		var err error
 		uri, out, err = a.runCmd(md)
 
+		if isFatalError(err) {
```
zerok, line 156: LGTM 🙂 Is there perhaps a way to cover this behaviour with a test?

zerok (approved):

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
dblinkhorn: thanks @zerok . added unit test for isFatalError.

PR #653, opened 2024-12-16
fix(techdocs-rewrite-relative-links): run `go mod tidy`

iainlane:
The last commit broke `techdocs-rewrite-relative-links`, leaving it in a state where it needs `go mod tidy`ing. Here we do that.

```
go: updates to go.mod needed; to update it:
	go mod tidy

```


paulajulve (approved):
it hit me too, took me a hot minute to understand what was going on

PR #663, opened 2024-12-20
print workflow uri immediately even when using --wait flag

dblinkhorn:


actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
```diff
@@ -58,41 +58,59 @@ func (a App) env() []string {
...
-	if err != nil {
-		return "", "", err
+	if err := cmd.Start(); err != nil {
+		return "", "", fmt.Errorf("failed to start command: %w", err)
 	}
 
-	uri, output := a.outputWithURI(cmdOutput)
+	uri, out, scanErr := a.outputWithURI(stdoutPipe)
+	if scanErr != nil {
+		return uri, out, scanErr
```
zerok, line 106: Please garbage collect the cmd and the pipe to avoid zombie processes 🙂

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
```diff
@@ -58,41 +58,59 @@ func (a App) env() []string {
...
+	var outputBuilder strings.Builder
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		outputBuilder.WriteString(line + "\n")
 
-	if len(matches) != 2 {
-		a.logger.Warn("Couldn't find workflow name in output - can't construct URI for launched workflow")
-		return "", output
+		matches := nameRe.FindStringSubmatch(line)
```
zerok, line 71: Even the matching can be skipped if `uri` is already set, right?

zerok (changes requested):
Looks good 🙂  Just some minor requests.

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
dblinkhorn: Thanks for pointing this out. I added a defer function for cleanup. Also, please feel free to nitpick all my PRs since I am very new to Go.

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
dblinkhorn: Updated the flow to address this

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/argo.go
zerok: Will do :) Regarding the defer: This is now also called in the happy path. IMO it should be enough to just do

```
_ = stdoutPipe.Close()
_ = cmd.Wait()
```

in the scanErr case and that's it 🙂

dblinkhorn: Updated, thanks!

zerok (approved):

PR #670, opened 2025-01-03
fix: login to GAR only on push events

dsotirakis:
Uses login-to-gar action only on push events, to allow PRs that come from forks to at least be able to build docker images on PRs.


npazosmendez (approved):
LGTM

iainlane:
@dsotirakis I'd consider adding a way to always log in. A time when you need to log in when not pushing is when your Dockerfile has another private image as a `FROM`.

PR #678, opened 2025-01-07
chore: remove `syft-sbom-report`

dsotirakis:
Removes `syft-sbom-report` as it's not used and doesn't have to be abstracted, at least for now :)


iainlane (approved):

clementduveau:
Hey, we had a customer asking why we stopped providing the SBOM. My guess is I can generate one manually reusing this action right ?

dsotirakis:
> Hey, we had a customer asking why we stopped providing the SBOM. My guess is I can generate one manually reusing this action right ?

Hey @clementduveau, we haven't stopped providing the SBOM, we believed that we didn't need a reusable action to do so as users could just use the sbom action straight without going through this repository.

PR #720, opened 2025-01-27
feat: add `custom-service-account` input to `*-to-gcs` actions

dsotirakis:
Adds `custom-service-account` input to `*-to-gcs` actions in case we need to login with a different service account rather than the one generated from tf code. This only works if a bucket is different than the default.

Here's a breakdown of how it works:
```
1. ${{ inputs.bucket && inputs.custom-service-account }}:

This part checks if inputs.bucket is not empty or null.
* If inputs.bucket is not empty, it evaluates to inputs.custom-service-account.
* If inputs.bucket is empty, it evaluates to false.

2. || '':

* This part is a logical OR operation.
* If the left side of the OR operation (inputs.bucket && inputs.custom-service-account) is false (i.e., inputs.bucket is empty), it evaluates to the right side, which is an empty string ''.
```

### DIFF
--- a/actions/login-to-gcs/README.md
+++ b/actions/login-to-gcs/README.md
@@ -39,10 +39,11 @@ $ gcloud storage cp OBJECT_LOCATION gs://DESTINATION_BUCKET_NAME
 
 ## Inputs
 
-| Name          | Type   | Description                                                                                |
-| ------------- | ------ | ------------------------------------------------------------------------------------------ |
-| `bucket`      | String | Name of bucket to upload to. Will default to grafanalabs-${repository.name}-${environment} |
-| `environment` | String | Environment for pushing artifacts (can be either dev or prod).                             |
+| Name                     | Type   | Description                                                                                |
+| ------------------------ | ------ | ------------------------------------------------------------------------------------------ |
+| `bucket`                 | String | Name of bucket to upload to. Will default to grafanalabs-${repository.name}-${environment} |
+| `environment`            | String | Environment for pushing artifacts (can be either dev or prod).                             |
+| `custom_service_account` | String | Custom service account to use for authentication.                                          |
 
 ## Outputs
 
--- a/actions/push-to-gcs/README.md
+++ b/actions/push-to-gcs/README.md
@@ -113,15 +113,16 @@ jobs:
 
 ## Inputs
 
-| Name            | Type   | Description                                                                                                                                                                                  |
-| --------------- | ------ | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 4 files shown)


actions/login-to-gcs/README.md
```diff
@@ -39,10 +39,11 @@ $ gcloud storage cp OBJECT_LOCATION gs://DESTINATION_BUCKET_NAME
...
 
-| Name          | Type   | Description                                                                                |
-| ------------- | ------ | ------------------------------------------------------------------------------------------ |
-| `bucket`      | String | Name of bucket to upload to. Will default to grafanalabs-${repository.name}-${environment} |
-| `environment` | String | Environment for pushing artifacts (can be either dev or prod).                             |
+| Name                     | Type   | Description                                                                                |
+| ------------------------ | ------ | ------------------------------------------------------------------------------------------ |
+| `bucket`                 | String | Name of bucket to upload to. Will default to grafanalabs-${repository.name}-${environment} |
+| `environment`            | String | Environment for pushing artifacts (can be either dev or prod).                             |
+| `custom_service_account` | String | Custom service account to use for authentication.                                          |
```
zerok, line 46: I think it would be better to name that field `service_account` (including its default value) as it would otherwise clash with the `bucket` field.

actions/push-to-gcs/README.md
```diff
@@ -113,15 +113,16 @@ jobs:
...
+| `bucket`                 | String | (Required) Name of bucket to upload to. Can be gathered from `login-to-gcs` action.                                                                                                          |
+| `path`                   | String | (Required) The path to a file or folder inside the action's filesystem that should be uploaded to the bucket. You can specify either the absolute path or the relative path from the action. |
+| `bucket_path`            | String | Bucket path where objects will be uploaded. Default is the bucket root.                                                                                                                      |
+| `environment`            | String | Environment for pushing artifacts (can be either dev or prod).                                                                                                                               |
+| `custom_service_account` | String | Custom service account to use for authentication. Used only when bucket input is not empty                                                                                                   |
```
zerok, line 122: Same as above, please remove the `custom_` prefix and hint at a default value in the docs 🙂

actions/push-to-gcs/action.yaml
```diff
@@ -69,6 +73,7 @@ runs:
       with:
         bucket: ${{ inputs.bucket }}
         environment: ${{ inputs.environment }}
+        custom-service-account: ${{ inputs.bucket && inputs.custom-service-account || '' }}
```
zerok, line 76: `-` vs `_` here.

zerok (approved):
🚀

PR #738, opened 2025-01-31
docs(check-drone-sig): init doc for workflow

zzehring:
This change introduces documentation for the `check-drone-signature` reusable workflow. It gives an overview for what the purpose of the check is for and how to set up and use the workflow.

### DIFF
--- a/.github/workflows/check-drone-signature.md
+++ b/.github/workflows/check-drone-signature.md
@@ -3,7 +3,7 @@
 ## Overview
 
 This document describes the intention and usage of the `check-drone-signature` GitHub Actions workflow.
-You can find [the workflow file at `.github/workflows/check-drone-signature.yml`][wf-file].
+You can find [the `check-drone-signature` workflow file here][wf-file].
 
 This reusable workflow checks [the signature of a Drone CI file][drone-sig] against our Drone server and fails if the signature is invalid.
 The check can help prevent against merging incorrectly signed files, which often occurs when the Drone CI config is modified without re-signing the file.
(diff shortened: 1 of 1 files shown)


.github/workflows/check-drone-signature.md
```diff
@@ -0,0 +1,60 @@
+# Check Drone Signature
+
+## Overview
+
+This document describes the intention and usage of the `check-drone-signature` GitHub Actions workflow.
+You can find [the workflow file at `.github/workflows/check-drone-signature.yml`][wf-file].
```
iainlane, line 6: This bit seems redundant since it's in the filename and if you followed a link it'd be in the link name, but keeping the ref to the workflow itself might be useful. 🙂

iainlane, line 6:
```suggestion
You can find [the workflow file at `.github/workflows/check-drone-signature.yaml`][wf-file].
```

.github/workflows/check-drone-signature.md
```diff
@@ -0,0 +1,60 @@
...
+  drone-signature-check:
+    uses: grafana/shared-workflows/.github/workflows/check-drone-signature.yaml@main
+    with:
+      drone_config_path: .drone.yml
+```
+
+This workflow file expects the following:
+
+* `main` is the name of your primary branch
+* `.drone.yml` is the filename for your Drone CI configuration file and it's located at the root of your repository
```
iainlane, line 51: In our other documentation we have this in a table in an "Inputs" section. It'd be good to do the same just for consistency.

iainlane (approved):
nice! just left a couple of nits but no need for re approval 👍

.github/workflows/check-drone-signature.md
zzehring: Removed explicit text link.

PR #795, opened 2025-02-20
feat(push-to-gar-docker): add custom label input

ahampal:
## Changes
Add optional field to specify custom labels that will be passed as args for the build-push-action.


ahampal:
> just to keep it consistent! looks good to me otherwise!

Updated with suggestions and fixed failing lint

dsotirakis (approved):

PR #821, opened 2025-02-27
fix: ensure every action disables git credential persistence

jdbaldry:
Reduces the possibility of inadvertently leaking the credentials in consuming workflows.

This has the potential to break workflows that rely on the persisted credentials and I haven't checked which ones may be likely to break

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ will ensure actions in this repo are always used at the same commit. To do this:
     # action), so if multiple actions check `shared-workflows` out, they don't
     # overwrite each other
     path: _shared-workflows-your-action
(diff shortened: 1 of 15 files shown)


svennergr (approved):

dsotirakis (approved):

zerok:
TIL https://github.com/actions/checkout/issues/485

zerok (approved):

jdbaldry:
Thanks both for reviewing, I can't merge this so please consider merging when you want :)

PR #846, opened 2025-03-12
feat: use auth_token in login-to-gar action

dsotirakis:
Instead of installing `gcloud` and then configuring docker manually (`gcloud auth configure-docker`), we can use `auth_token` to login to the docker registry.

Working example:

image
([workflow](https://github.com/grafana/dsotirakis-test-repo/actions/runs/13812420383/job/38636912634) - the error comes from the service account auth try and is expected)


iainlane (approved):
love this!

PR #895, opened 2025-04-08
fix: update buildx version to `latest`

dsotirakis:
Due to [decommissioned cache service brownouts](https://github.blog/changelog/2025-03-20-notification-of-upcoming-breaking-changes-in-github-actions/#decommissioned-cache-service-brownouts), we need to update and pin the buildx version (see comment [here](https://github.com/docker/build-push-action/issues/1345#issuecomment-2770572479)) to prevent our actions from failing. 

Working example:
https://github.com/grafana/website/actions/runs/14350268712/job/40227545764?pr=25031#step:21:1

### DIFF
--- a/actions/push-to-gar-docker/action.yaml
+++ b/actions/push-to-gar-docker/action.yaml
@@ -153,7 +153,7 @@ runs:
       uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
       with:
         driver: ${{ inputs.docker-buildx-driver }}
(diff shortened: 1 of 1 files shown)


jdbaldry:
FYI, this is now full blackout as of today.


> Error: buildx failed with: ERROR: failed to solve: This legacy service is shutting down, effective April 15, 2025. Migrate to the new service ASAP. For more information: https://gh.io/gha-cache-sunset

dsotirakis:
> just a qq, should we use `latest` or set up Renovate to track this?

can do, since that's the comment in the GH issue I linked above says. would be better if we used Renovate, I'll create an issue

iainlane (approved):

PR #904, opened 2025-04-16
feat(push-to-gar-docker): add target input for multi-stage builds

ricky-undeadcoders:
Expose the underlying `target` input from docker's [build-push-action](https://github.com/docker/build-push-action/blob/master/action.yml#L104)


dblinkhorn (approved):

zerok (approved):
Just out of curiousity: Why can the original input be used by other actions? That value has to come from somewhere 🙂

zerok:
> Just out of curiousity: Why can the original input be used by other actions? That value has to come from somewhere 🙂

nvm, I'm stupid :)

PR #905, opened 2025-04-17
ci: don't persist shared workflows folder after action is done

dsotirakis:
When CI jobs are running git status after using `push-to-gar-docker` or `push-to-gcs` actions, the checked out `shared-workflows` directory appears. This breaks some CI check workflows, where the expectation is that there's not going to be any diff after certain operations (like building npm, or linting etc.). 

This PR ensures that at the end of each the above two actions, the directory will be cleaned up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,13 @@ will ensure actions in this repo are always used at the same commit. To do this:
   uses: ./_shared-workflows-your-action/actions/some-action
   with:
     some-input: some-value
+
+# This step ensures the checkout directory is cleaned up even if previous steps fail
+- name: Cleanup checkout directory
+  if: ${{ !cancelled() }}
+  shell: bash
+  run: |
+    rm -rf _shared-workflows-your-action
 ```
 
 ### Use separate files for shell scripts so they're linted
@@ -127,7 +134,7 @@ Also, the following block should be added in the README file which will be respo
 
 `README.md`:
 
-````markdown
+```markdown
 # my-new-action
 
 This is my new action which does awesome stuff!
@@ -149,6 +156,3 @@ jobs:
 ```
 
 <!-- x-release-please-end-version -->
-````
-
-Every semver-like string between the `x-release-please-start-version` and `x-release-please-end-version` comments will be updated with the new version whenever the component is released.
--- a/README.md
+++ b/README.md
@@ -69,11 +69,16 @@ will ensure actions in this repo are always used at the same commit. To do this:
     some-input: some-value
 
 # This step ensures the checkout directory is cleaned up even if previous steps fail
-- name: Cleanup checkout directory
-  if: ${{ !cancelled() }}
-  shell: bash
-  run: |
-    rm -rf _shared-workflows-your-action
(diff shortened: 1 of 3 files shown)


actions/push-to-gar-docker/action.yaml
```diff
@@ -170,3 +170,9 @@ runs:
         platforms: ${{ inputs.platforms }}
         ssh: ${{ inputs.ssh }}
         build-contexts: ${{ inputs.build-contexts }}
+
+    - name: Cleanup checkout directory
+      if: always()
```
iainlane, line 175:
```suggestion
      if: ${{ !cancelled() }}
```

From [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always). `always()` has caused us problems with stuck jobs before - it gets in the way of cancellation...

actions/push-to-gar-docker/action.yaml
```diff
@@ -170,3 +170,9 @@ runs:
...
+      shell: bash
+      run: |
+        rm -rf _shared-workflows-push-to-gar
```
iainlane, line 178: since this is going to run all the time it might be an idea to be defensive

```suggestion
      run: |
        # Check that the directory looks OK before removing it
        if ! [ -d "_shared-workflows-push-to-gar/.git" ]; then
          echo "::warning Not removing shared workflows directory: doesn't look like a git repository"
          exit 0
        fi

        rm -rf _shared-workflows-push-to-gar
```

iainlane (commented):
Nice 👍 . Could you update [the example in the readme](https://github.com/grafana/shared-workflows/blob/801040de7886465a40d713511ae5fdb35a5f36f1/README.md?plain=1#L43) with this step please? 🙏

README.md
```diff
@@ -147,8 +154,6 @@ jobs:
       - id: do-stuff
         uses: grafana/shared-workflows/actions/my-new-action@my-new-action-v1.0.0
 ```
-
-<!-- x-release-please-end-version -->
 ````
 
-Every semver-like string between the `x-release-please-start-version` and `x-release-please-end-version` comments will be updated with the new version whenever the component is released.
+<!-- x-release-please-end-version -->
```
iainlane, line 159: This should stay in the visible bit no?

README.md
```diff
@@ -67,6 +67,13 @@ will ensure actions in this repo are always used at the same commit. To do this:
   uses: ./_shared-workflows-your-action/actions/some-action
   with:
     some-input: some-value
+
+# This step ensures the checkout directory is cleaned up even if previous steps fail
+- name: Cleanup checkout directory
+  if: ${{ !cancelled() }}
+  shell: bash
+  run: |
+    rm -rf _shared-workflows-your-action
```
iainlane, line 76: Probably sync this with the code we're using

iainlane (approved):
yessss

README.md
dsotirakis: Oh docs got messed up. Yes, gonna revert that change

PR #605, opened 2024-12-02
feat(zizmor): add a reusable workflow for zizmor GitHub Actions static analysis

iainlane:
_This depends on #930._

This is a [reusable workflow] which runs the [`zizmor`][zizmor] static analysis tool on a repo's GitHub Actions workflow files. This will report things such as whether there is potential for untrusted code to be injected via a template. See a full list of checks in [the documentation][zizmor-checks].

This workflow will run zizmor, upload results to GitHub's code scanning service (requires an Advanced Security subscription for private repositories), and comment on the pull request with the results. The comment will be re-posted on each run - and previous comments hidden - so the most recent comment will always show the current results.

[reusable workflow]: https://docs.github.com/en/actions/using-workflows/reusing-workflows
[zizmor]: https://woodruffw.github.io/zizmor/
[zizmor-checks]: https://woodruffw.github.io/zizmor/audits/

PR #930, opened 2025-04-28
fix(everything): fix all things for zizmor

iainlane:
Zizmor is reporting a _lot_ of problems here. It's mostly input sanitisation that is missing (indirecting variables through the environment), and setting permissions.

This fixes them all, so that the new Zizmor workflow is clean.

Prerequisite of #605 (adding a reusable workflow to call Zizmor). This makes the "self test" in there pass.

### DIFF
--- a/actions/generate-openapi-clients/action.yaml
+++ b/actions/generate-openapi-clients/action.yaml
@@ -26,7 +26,7 @@ inputs:
     required: true
   modify-spec-script:
     description: |
-      The path to an executable script that modifies the OpenAPI spec before generating the client. 
+      The path to an executable script that modifies the OpenAPI spec before generating the client.
       The spec will be piped into the script and the script should output the modified spec to stdout.
       Note: This is used as a workaround for the OpenAPI generator not supporting certain features. By using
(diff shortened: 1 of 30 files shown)


iainlane:
I've not set auto-merge but feel free to merge this and also push any fixes if you like.

We will need to fix Renovate first though.

jamesc-grafana (approved):

actions/generate-openapi-clients/action.yaml
```diff
@@ -36,50 +36,58 @@ inputs:
...
-      env:
-        SPEC_PATH: ${{ inputs.spec-path }}
 
     # Generate the client
     - shell: bash
       run: ${GITHUB_ACTION_PATH}/generate.sh
       env:
         OUTPUT_DIR: ${{ inputs.output-dir }}
         PACKAGE_NAME: ${{ inputs.package-name }}
+        SPEC_PATH: ${{ steps.modify-spec.outputs.SPEC_PATH }}
```
svennergr, line 80:
```suggestion
        SPEC_PATH: ${{ steps.modify-spec.outputs.spec-path }}
```

PR #931, opened 2025-04-28
fix(get-vault-secrets): add outputs for export_env false

guicaulada:
The option to set export_env to false was given, but there was no way to retrieve the secrets when that was set.

### DIFF
--- a/actions/get-vault-secrets/README.md
+++ b/actions/get-vault-secrets/README.md
@@ -38,11 +38,11 @@ jobs:
             ENVVAR2=test-secret:key1
 
       # Use the secrets
-      # You can use the envvars directly in scripts or use the `${{ env.* }}` accessor in the workflow
+      # You can use the envvars directly in scripts
       - name: echo
         run: |
           echo "$ENVVAR1"
-          echo "${{ env.ENVVAR2 }}"
+          echo "${ENVVAR2}"
 ```
 
 <!-- x-release-please-end-version -->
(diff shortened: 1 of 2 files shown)


actions/get-vault-secrets/action.yaml
```diff
@@ -32,6 +32,11 @@ inputs:
       Whether to export secrets as environment variables, making them available to all subsequent steps. Defaults to `true`.
     default: "true"
 
+outputs:
+  secrets:
+    description: "JSON object containing all the secrets"
+    value: ${{ steps.import-secrets.outputs.secrets }}
```
MattIPv4, line 38:
```suggestion
    value: ${{ toJson(steps.import-secrets.outputs) }}
```

MattIPv4:
Would it be worth adding an example of using this output (with or without `export_env: false`) to the README?

actions/get-vault-secrets/README.md
```diff
@@ -44,3 +46,46 @@ jobs:
 ```
 
+
+### Using Outputs
+
+You can also use the action with `export_env: false` to get secrets as outputs instead of environment variables:
+
+```yaml
```
MattIPv4, line 54: Does this need to be within `x-release-please-start-version` like the example above?

MattIPv4 (approved):
Functionally the output works

actions/get-vault-secrets/README.md
```diff
@@ -44,3 +46,46 @@ jobs:
...
+          export_env: false
+
+      # Use the secrets from the JSON output in the env block
+      - name: echo
+        env:
+          SECRET1: ${{ fromJSON(steps.get-secrets.outputs.secrets).SECRET1 }}
+          SECRET2: ${{ fromJSON(steps.get-secrets.outputs.secrets).SECRET2 }}
+        run: |
+          echo "$SECRET1"
+          echo "${{ env.SECRET2 }}"
```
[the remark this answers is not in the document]
guicaulada: True! I will remove it from both examples.

actions/get-vault-secrets/README.md
guicaulada: Good catch! Added.

dblinkhorn (approved):

PR #944, opened 2025-04-29
feat: create version of zizmor workflow to be used as a ruleset

jamesc-grafana:
We want to use Zizmor as an organization wide ruleset, to do this we need to have a workflow that is in a publicly accessible repository (like this one). The current options for Zizmor in this repo are the `reusable-zizmor` which we can't use for this use case because rulesets require the workflow to be defined as triggering on `pull_request` or `push`. The other option is to use the `self-zizmor` which utilises the CodeQL integration, which we don't want to use in the current iteration.

Due to the above reasoning, we have created an almost identical action to the `self-zizmor` that just disables the `codeql-enabled` input so that we don't upload the SARIF results to GitHub at the moment

### DIFF
--- a/.github/workflows/rule-zizmor.yml
+++ b/.github/workflows/rule-zizmor.yml
@@ -1,11 +1,11 @@
-name: zizmor GitHub Actions static analysis
+name: Run Zizmor Ruleset
 on:
   push:
   pull_request:
 
 jobs:
   zizmor:
-    name: Run zizmor from current branch (self test)
+    name: Zizmor Ruleset
 
     permissions:
       actions: read
--- a/.github/workflows/rule-zizmor.yml
+++ b/.github/workflows/rule-zizmor.yml
@@ -1,0 +1,19 @@
+name: zizmor GitHub Actions static analysis
+on:
+  push:
+  pull_request:
+
+jobs:
+  zizmor:
+    name: Run zizmor from current branch (self test)
+
+    permissions:
+      actions: read
+      contents: read
+
+      pull-requests: write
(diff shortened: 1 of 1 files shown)


iainlane:
How come we don't want the CodeQL thing?

iainlane:
Just thinking quickly about this...

I think for an org-wide action we should support supplying a default a config file, so we can tweak things centrally. But I would defer to one in the repo if it exists.

iainlane:
> > How come we don't want the CodeQL thing?
> 
> We're still ironing out licensing for Advanced Security on private repos I believe, so in order to blanket roll this out, we need to be able to run without it first

Ah. Does it fail to upload? I seemed to remember that it was a warning... I'll try this in a second.

If it _does_ fail, what do you think about adding `continue-on-error: true` to that step instead? What I'm thinking is that it'd be great to have this automatically work for contexts where it can, because what you get is pretty nice.

jamesc-grafana:
> > > How come we don't want the CodeQL thing?
> > 
> > 
> > We're still ironing out licensing for Advanced Security on private repos I believe, so in order to blanket roll this out, we need to be able to run without it first
> 
> Ah. Does it fail to upload? I seemed to remember that it was a warning... I'll try this in a second.
> 
> If it _does_ fail, what do you think about adding `continue-on-error: true` to that step instead? What I'm thinking is that it'd be great to have this automatically work for contexts where it can, because what you get is pretty nice.

Yeah that sounds like a better option, then I can reuse the `self-zizmor` action for the ruleset

PR #946, opened 2025-04-29
feat(zizmor): use a default config file

iainlane:
If the target repo hasn't got its own config file, we'll use our defaults. In this way we can provide defaults for our users but let them be overridden.

Note that this is inline in the CI job itself. I don't think that's ideal but I didn't find a better way to do this. It doesn't seem possible for reusable workflows to find out their own ref, which is what they'd need to do to get a sibling file from their repo.

I've made a support ticket with GitHub about that. If they give me a fix I'll implement that.

### DIFF
--- a/.github/workflows/reusable-zizmor.yml
+++ b/.github/workflows/reusable-zizmor.yml
@@ -27,6 +27,27 @@ on:
         type: string
         default: "ubuntu-latest"
 
+      # TODO: This should _not_ be inline. It should be a file like
+      # `.github/zizmor.yml` alongside the reusable workflow. But
+      # unfortunately we didn't find a way to load such a file so far.
+      default-config:
+        description: The default configuration to use.
+        required: false
+        type: string
+        default: |
+          rules:
+            unpinned-uses:
+              config:
+                policies:
+                  actions/*: any # trust GitHub
+                  grafana/*: any # trust Grafana
+
+      always-use-default-config:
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-zizmor.yml
```diff
@@ -60,6 +60,38 @@ jobs:
...
+            # No action needed, zizmor will find it
+            echo "Using zizmor.yml found in repository root."
+            exit 0
+          fi
+
+          if [ -f ".github/zizmor.yml" ]; then
+            # No action needed, zizmor will find it
+            echo "Using .github/zizmor.yml found in repository."
+            exit 0
+          fi
```
jamesc-grafana, line 78: Could we have an input called `prefer-local-config` which defaults to true? I imagine there will be circumstances where we'd prefer to enforce a central config over user specified ones

iainlane: Alright, can do.

I'm thinking that I'll move this to an input too, so that we can have the "central" workflow able to control the default if/when that ends up living somewhere else.

iainlane: OK - take a look at this one 👍

jamesc-grafana (approved):

iainlane:
oh HEH it's got a conflict, I didn't notice! sorry!

PR #948, opened 2025-04-30
feat(zizmor): always try to upload results to CodeQL

iainlane:
This might not work if the repository isn't eligible (it's a paid feature for private repos), but we can try and not worry about it if it doesn't.


iainlane:
https://github.com/grafana/shared-workflows/pull/947

PR #951, opened 2025-04-30
fix(zizmor): github token as input

jamesc-grafana:
This PR enables the passing in of a GitHub token into the workflow from a parent, this will enable private actions to be read when we use an application to authenticate with the GitHub API

### DIFF
--- a/.github/workflows/reusable-zizmor.yml
+++ b/.github/workflows/reusable-zizmor.yml
@@ -48,6 +48,12 @@ on:
         type: boolean
         default: false
 
+      github-token:
+        description: "The GitHub token to use for the job"
+        required: false
+        type: string
+        default: ${{ github.token }}
+
 permissions: {}
 
 jobs:
@@ -126,7 +132,7 @@ jobs:
 
       - name: Run zizmor
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ inputs.github-token }}
           ZIZMOR_CONFIG: ${{ steps.setup-config.outputs.zizmor-config }}
         shell: sh
         run: >-


.github/workflows/reusable-zizmor.yml
```diff
@@ -126,14 +131,15 @@ jobs:
 
       - name: Run zizmor
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ inputs.github-token || github.token }}
```
iainlane, line 134: Let's do this by setting `${{ github.token }}` as the default value of the input

.github/workflows/reusable-zizmor.yml
```diff
@@ -126,14 +131,15 @@ jobs:
...
           ZIZMOR_CONFIG: ${{ steps.setup-config.outputs.zizmor-config }}
         shell: sh
         run: >-
           zizmor
           --format sarif
           --min-severity "${MIN_SEVERITY}"
           --min-confidence "${MIN_CONFIDENCE}"
+          --gh-token="${GH_TOKEN}"
```
iainlane, line 142: Don't need both the env var and the flag - pick one!

iainlane (approved):
Cheers. I think we need this!

One thing. If we're going to be passing in a token, that will be a single rate limit for the whole org (15,000 calls per hour). For a job that runs on everything I have a feeling we'll hit that and then get in trouble.

One thing that might help is that I noticed [Zizmor supports a cache](https://woodruffw.github.io/zizmor/usage/#caching-between-runs). We might wire that up?

.github/workflows/reusable-zizmor.yml
jamesc-grafana: I did try this, however, because the permissions are scoped to the job itself, setting it as the input's default actually results in an empty token being set

.github/workflows/reusable-zizmor.yml
jamesc-grafana: Sure, will remove the flag

.github/workflows/reusable-zizmor.yml
iainlane: Ah, yes, that makes sense! Would it work at the job level?

If so, might be slightly nicer, but that's a nit at this point.

.github/workflows/reusable-zizmor.yml
```diff
@@ -124,16 +129,24 @@ jobs:
           --version "${ZIZMOR_VERSION}"
           zizmor
 
+      - name: Zizmor cache
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4
+        with:
+          path: ${{ runner.temp }}/.cache/zizmor
+          key: zizmor-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('**/.github/workflows/*', '**/action.{yaml,yml}') }}
```
iainlane, line 136: Hmm, I think this means we'll throw the cache away whenever the actions change.

```suggestion
          key: zizmor-${{ runner.os }}-${{ runner.arch }}
```

I wonder what the impact of having a more general cache key would be? Feels to me like previously cached values would still be valid - it'll be requests/responses made to the GitHub API and those should be good to be cached even if the source of them is different.

iainlane (approved):
Cheers for this. Looks good 🌈 

See my comment about the cache key. Looks to me like we can use the cache a bit more than we would do. That can be fixed now or in a followup.

.github/workflows/reusable-zizmor.yml
jamesc-grafana: That makes a lot of sense

.github/workflows/reusable-zizmor.yml
jamesc-grafana: Yeah that's much cleaner than the previous version

PR #952, opened 2025-04-30
perf: simplify zizmor workflow

erj826:
Use the pre-build python wheel for Zizmor to avoid additional overhead before running.


erj826:
> I think that's a good idea to do. The reason we're building is because a previous version of the workflow needed a specific commit and we had to build zizmor ourselves to get that. That's not the case any more, so I don't see a problem with using the binaries. It should save a lot on the bulid time.
> 
> Before we can merge this, can you please remove all of the other changes and keep this to the minimal stuff required to install using `uv`? If you have other improvemnets to contribute, that's great - just let's do them in focused PRs each implementing one change.

Sure thing! I'll open a new PR

PR #953, opened 2025-05-01
feat(reusable-zizmor): store default config file separately

iainlane:
We've got a TODO item saying that storing the config in the YAML definition of the reusable workflow itself isn't the best. It's hard to find, edit, and there's no linting since it's considered a string and not YAML itself.

It's not super obvious how to do this, though. What we want to do is use a file which is alongside the workflow. So if we're called like:

```yaml
job:
  zizmor:
    uses: grafana/shared-workflows/.github/workflows/reusable-zizmor.yml@abc123
```

...the correct thing to do is to use the file from `abc123` and not anything else which might be incompatible or change independently of the caller updating their pointer to us.

GitHub doesn't give us a checkout of the reusable workflow. So we need to fetch the file from GitHub ourselves. What we need to do that is the ref (`abc123`). Unfortunately this isn't easily available. For example it's not in the `${{ github }}` context, or any of the predefined environment variables.

But one place it _is_ available is [in the OIDC ID token][oidc-reusable].

Given that, by fetching the `job_workflow_ref` from the ID token, we can find out our current reference and retrieve the default config file from the repository at that ref.

This lets us store it separately, rather than having it inline where it's much harder to edit.

Here we do that - we fetch the ID token, convert it to a SHA if it's a named ref like `refs/heads/main`, and then fetch `.github/zizmor.yml` from this repository at that SHA. The file is cached, so we don't have re-fetch the file every time.

[oidc-reusable]: https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/using-openid-connect-with-reusable-workflows

### DIFF
--- a/.github/workflows/reusable-zizmor.md
+++ b/.github/workflows/reusable-zizmor.md
@@ -35,6 +35,10 @@ jobs:
       actions: read
       contents: read
 
+      # used in the `job-workflow-ref` job to fetch an OIDC token, which allows
+      # the run to determine its ref
+      id-token: write
+
       # required to comment on pull requests with the results of the check
       pull-requests: write
       # required to upload the results to GitHub's code scanning service
--- a/.github/workflows/reusable-zizmor.yml
+++ b/.github/workflows/reusable-zizmor.yml
@@ -27,21 +27,6 @@ on:
         type: string
         default: "ubuntu-latest"
 
-      # TODO: This should _not_ be inline. It should be a file like
-      # `.github/zizmor.yml` alongside the reusable workflow. But
-      # unfortunately we didn't find a way to load such a file so far.
-      default-config:
-        description: The default configuration to use.
-        required: false
-        type: string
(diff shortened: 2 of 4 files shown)


guicaulada (commented):
Would this still allow other repositories to set their own zizmor.yml config following the documentation [here](https://woodruffw.github.io/zizmor/configuration/)?

jamesc-grafana (approved):
LGTM, I wonder whether we include the reviewdog config to ban the use of that here? 🤔

jamesc-grafana:
> Would this still allow other repositories to set their own zizmor.yml config following the documentation [here](https://woodruffw.github.io/zizmor/configuration/)?

The `setup-config` step will exit before a `ZIZMOR_CONFIG` path is set. Zizmor tries the root of the directory and `.github/zizmor.yml` by default and only looks at the runner temp if it's set

PR #954, opened 2025-05-01
perf: use prebuilt zizmor wheel

erj826:
Use the pre-built zizmor wheel to avoid building the binary on the first run.

If we see a future where we want to build the Zizmor binary from a commit hash instead of a pre-built wheel it's probably easier to close this PR and continue to use the Rust toolchain.

### DIFF
--- a/.github/workflows/reusable-zizmor.yml
+++ b/.github/workflows/reusable-zizmor.yml
@@ -121,7 +121,7 @@ jobs:
           enable-cache: true
           activate-environment: true
           cache-suffix: ${{ env.ZIZMOR_VERSION }}
-          cache-dependency-glob: ''
+          cache-dependency-glob: ""
 
       - name: Install Zizmor
         shell: bash
--- a/.github/workflows/reusable-zizmor.yml
+++ b/.github/workflows/reusable-zizmor.yml
@@ -113,22 +113,18 @@ jobs:
           EOC
           echo "zizmor-config=${ZIZMOR_CONFIG}" | tee -a "${GITHUB_OUTPUT}"
 
-      - name: Setup Rust
-        uses: actions-rust-lang/setup-rust-toolchain@9d7e65c320fdb52dcd45ffaa68deb6c02c8754d9 # v1.12.0
-        with:
-          cache-on-failure: true
+      - uses: astral-sh/setup-uv@c7f87aa956e4c323abf06d5dec078e358f6b4d04 # v6.0.0
 
-      - name: Get zizmor
-        shell: sh
+      - name: 🐍 Create UV venv & install Zizmor
+        shell: bash
         env:
-          # renovate: datasource=github-tags depName=woodruffw/zizmor
+          # renovate: datasource=pypi depName=zizmor
           ZIZMOR_VERSION: 1.6.0
-        run: >-
-          RUSTFLAGS=-Awarnings
-          cargo
-          install
-          --version "${ZIZMOR_VERSION}"
-          zizmor
+        run: |
+          echo "🐣 Creating UV virtualenv"
+          uv venv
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-zizmor.yml
```diff
@@ -113,22 +113,18 @@ jobs:
...
 
-      - name: Setup Rust
-        uses: actions-rust-lang/setup-rust-toolchain@9d7e65c320fdb52dcd45ffaa68deb6c02c8754d9 # v1.12.0
-        with:
-          cache-on-failure: true
+      - uses: astral-sh/setup-uv@c7f87aa956e4c323abf06d5dec078e358f6b4d04 # v6.0.0
 
-      - name: Get zizmor
-        shell: sh
+      - name: Create UV venv & install Zizmor
```
iainlane, line 118: What do you think about using [`setup-uv`](https://github.com/astral-sh/setup-uv)?

```yaml
- name: Install the latest version of uv
  uses: astral-sh/setup-uv@6b9c6063abd6010835644d4c2e1bef4cf5cd0fca # v6.0.1
  with:
    enable-cache: true
```

I got this from [Zizmor's docs](https://woodruffw.github.io/zizmor/usage/#use-in-github-actions).

erj826: Updated the setup uv step to enable cache and use `activate-environment: true` to avoid manually creating the venv. Good call!

iainlane (commented):
Cheers. I think this is _nearly_ ready to go.

I noticed something in the logs. I wanted to check that caching of the zizmor binary was working properly. In there I can see

```
Warning: No file matched to [**/*requirements*.txt,**/*requirements*.in,**/*constraints*.txt,**/*constraints*.in,**/pyproject.toml,**/uv.lock]. The cache will never get invalidated. Make sure you have checked out the target repository and configured the cache-dependency-glob input correctly.
```

Which is sort of fine since we're not building our own Python project here. But it says "the cache will never get invalidated" which is actually an issue for us. If we bump `ZIZMOR_VERSION` then it _needs_ to be invalidated, or we could end up restoring the wrong version and using that or at least wasting a trip to the cache.

So how about this:

1. Move `ZIZMOR_VERSION` to an `env` key at the job level.
2. In the "Setup UV" step, set [`cache-suffix`](https://github.com/astral-sh/setup-uv#enable-caching) to `${ZIZMOR_VERSION}`.

Then we should have a proper cache key.

erj826:
> Cheers. I think this is _nearly_ ready to go.
> 
> I noticed something in the logs. I wanted to check that caching of the zizmor binary was working properly. In there I can see
> 
> ```
> Warning: No file matched to [**/*requirements*.txt,**/*requirements*.in,**/*constraints*.txt,**/*constraints*.in,**/pyproject.toml,**/uv.lock]. The cache will never get invalidated. Make sure you have checked out the target repository and configured the cache-dependency-glob input correctly.
> ```
> 
> Which is sort of fine since we're not building our own Python project here. But it says "the cache will never get invalidated" which is actually an issue for us. If we bump `ZIZMOR_VERSION` then it _needs_ to be invalidated, or we could end up restoring the wrong version and using that or at least wasting a trip to the cache.
> 
> So how about this:
> 
> 1. Move `ZIZMOR_VERSION` to an `env` key at the job level.
> 2. In the "Setup UV" step, set [`cache-suffix`](https://github.com/astral-sh/setup-uv#enable-caching) to `${ZIZMOR_VERSION}`.
> 
> Then we should have a proper cache key.

Good catch!

iainlane (approved):
Great, let's do this. Thanks! 🦄

Next, if you have time, can you propose an update to https://github.com/grafana/security-github-actions/blob/main/.github/workflows/self-zizmor.yaml to get this version used for the org's runs please?

PR #959, opened 2025-05-07
feat: adding new remove-checkout-credentials-action

zerok:
This action can be used to clean up Git credentials after an `actions/checkout` call in case `persist-credentials: false` is not an option.

### DIFF
--- a/.github/workflows/test-remove-checkout-credentials.yaml
+++ b/.github/workflows/test-remove-checkout-credentials.yaml
@@ -1,0 +1,45 @@
+name: Test remove-checkout-credentials action
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "actions/remove-checkout-credentials/**"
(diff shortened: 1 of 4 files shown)


.github/workflows/test-remove-checkout-credentials.yaml
```diff
@@ -0,0 +1,45 @@
+name: Test login-to-gar action
```
dsotirakis, line 1:
```suggestion
name: Test remove-checkout-credentials
```

dsotirakis: LGTM otherwise! Not approving yet to make sure we won't mess up with action test names 😁

zerok: Fixed, thanks :)

dsotirakis (approved):

PR #970, opened 2025-05-08
fix(reusable-zizmor): fix runs from forks

iainlane:
Forks can't get an ID token. We use this method to look up the version we're called at: get an OIDC ID token and then use the `job_workflow_ref` claim. We're currently hard failing when this doesn't work, which means we throw an error for runs from forks.

We do this whole ID token dance to find our current ref, so we can fetch the org-wide config. This is important but it's not _critical_ - not enough to fail the run. It would be better to succeed by falling back to using the default config from _upstream_.

### DIFF
--- a/.github/workflows/reusable-zizmor.md
+++ b/.github/workflows/reusable-zizmor.md
@@ -17,7 +17,7 @@ show the current results.
 
 ## Examples
 
-**Online Checks**
+### Online Checks
 
 ```yaml
 name: Zizmor GitHub Actions static analysis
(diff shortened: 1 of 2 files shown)


iainlane:
Here's some results

 - Failure (run against current main): https://github.com/grafana/shared-workflows/pull/971
 - All good (we found the error but we didn't fail any more): https://github.com/grafana/shared-workflows/actions/runs/14906940627?pr=972

jamesc-grafana (approved):
LGTM 🥳

PR #979, opened 2025-05-12
fix: only use zizmor config if it exists

jamesc-grafana:
We're finding that some forks of code cannot pull the content from this repo's zizmor config. We should ensure that zizmor runs with the default configuration if there is no configuration provided

### DIFF
--- a/.github/workflows/reusable-zizmor.yml
+++ b/.github/workflows/reusable-zizmor.yml
@@ -75,7 +75,6 @@ jobs:
       owner: ${{ steps.get-job-workflow-ref.outputs.owner }}
       repo: ${{ steps.get-job-workflow-ref.outputs.repo }}
       sha: ${{ steps.get-job-workflow-ref.outputs.sha }}
-      success: ${{ steps.get-job-workflow-ref.outputs.success }}
 
     steps:
       - id: setup-node
@@ -208,7 +207,6 @@ jobs:
               core.setOutput('owner', owner);
               core.setOutput('repo', repo);
               core.setOutput('sha', sha);
-              core.setOutput('success', 'true');
             } catch (error) {
               // On errors, we log an error messge, but we don't fail. It's
               // better to run with the default config than not run at all.
@@ -242,6 +240,7 @@ jobs:
       ZIZMOR_VERSION: 1.6.0
       GH_TOKEN: ${{ inputs.github-token || github.token }}
       ZIZMOR_EXTRA_ARGS: ${{ inputs.extra-args }}
+      DEFAULT_ZIZMOR_CONFIG_DOWNLOADED: ${{ needs.job-workflow-ref.outputs.sha }}
 
     steps:
       - name: Harden the runner (Audit all outbound calls)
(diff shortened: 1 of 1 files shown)


iainlane:
Can you share link to a log please? I'd like to understand how it's failing.

jamesc-grafana:
> Can you share link to a log please? I'd like to understand how it's failing.

Absolutely, it's failing in https://github.com/grafana/faro-web-sdk/actions/runs/14917848027/job/41907354221?pr=1201 for the reason:

```
Stack trace: Error: We're unable to look up the version of the Zizmor workflow being called, so we can't fetch the `grafana` default configuration. Zizmor's own default will be used. Is the `is-token: write` permission set? If so, is this a run from a fork? Unfortunately we're unable to do this lookup for pull requests from forks currently.
```

.github/workflows/reusable-zizmor.yml
```diff
@@ -348,8 +348,13 @@ jobs:
             echo "always-use-default-config is set. Using default config."
           fi
 
+          # Some forks are having issues with cloning the zizmor config file, so in the worst case
+          # we'll use the default configuration that's built in.
+
```
iainlane, line 353: We've already covered this extensively in the message above (line 107). _All_ forks will fail and fall back to the default. I'd drop this - we don't need to say again.

```suggestion
```

iainlane (commented):
> > Can you share link to a log please? I'd like to understand how it's failing.
> 
> Absolutely, it's failing in [grafana/faro-web-sdk/actions/runs/14917848027/job/41907354221?pr=1201](https://github.com/grafana/faro-web-sdk/actions/runs/14917848027/job/41907354221?pr=1201) for the reason:
> 
> ```
> Stack trace: Error: We're unable to look up the version of the Zizmor workflow being called, so we can't fetch the `grafana` default configuration. Zizmor's own default will be used. Is the `is-token: write` permission set? If so, is this a run from a fork? Unfortunately we're unable to do this lookup for pull requests from forks currently.
> ```

Thanks! Hmm, I see. I'm trying to make this be empty in the step where we run Zizmor. Currently it has

```
ZIZMOR_CONFIG: /home/runner/work/_temp/zizmor.yml
```

in the env, which is wrong.

If it's okay, I'd like to suggest a _very slightly_ different approach.

I've tried to be careful and make the `job-workflow-ref` job's outputs be empty when we can't do this query. I can see that they are in this case, which is good. I'd like it if we used this as our signal, so that if we fail to write the default config for any other reason we treat that as a bug that we need to fix - i.e. we get errors for it.

My suggestion is to do the following

- Make an env var (or an output of the `job-workflow-ref` job, whatever feels best) which indicates whether the fetch was able to happen or not. It can work like `FORCE_DEFAULT_CONFIG` works, and then the conditions could be something like `if: env.ZIZMOR_DEFAULT_CONFIG_DOWNLOADED` since `''` is falsy.
- Skip the `cache-config` and `fetch-config` steps in that case, since they won't work (they need to know the repo & sha)
- Keep an `if` check like you have in the PR currently, but make it something like

```bash
if [ -n "${ZIZMOR_DEFAULT_CONFIG_DOWNLOADED}" ]; then
   echo '...'
fi
```

Hope this makes sense. Thanks for the PR 👍

.github/workflows/reusable-zizmor.yml
```diff
@@ -207,6 +208,7 @@ jobs:
               core.setOutput('owner', owner);
               core.setOutput('repo', repo);
               core.setOutput('sha', sha);
+              core.setOutput('success', 'true');
```
iainlane, line 211: This shouldn't be needed: we can check for `sha` being empty

.github/workflows/reusable-zizmor.yml
```diff
@@ -327,6 +330,7 @@ jobs:
         id: setup-config
         env:
           FORCE_DEFAULT_CONFIG: ${{ inputs.always-use-default-config && 'true' || '' }}
+          DEFAULT_ZIZMOR_CONFIG_DOWNLOADED: ${{ needs.job-workflow-ref.outputs.success == 'true' && 'true' || '' }}
```
iainlane, line 333: This could be in the `env`, and then the other `if` conditions could reference it (they wouldn't need the `== 'true'` then)

iainlane (approved):
cheers. looks great 👍 

I left some optional suggestions.

One thing I would advise you to do before merging would be to explicitly check this from a non-`grafana` repository. You should be able to call from this branch, and we can make sure that it works!

.github/workflows/reusable-zizmor.yml
jamesc-grafana: Addressed in 8bc832f

.github/workflows/reusable-zizmor.yml
jamesc-grafana: Addressed in 8bc832f

jamesc-grafana:
> One thing I would advise you to do before merging would be to explicitly check this from a non-grafana repository. You should be able to call from this branch, and we can make sure that it works!

Tested in this PR against this branch: https://github.com/grafana/shared-workflows/pull/986

iainlane:
Do it (maybe Monday morning at this point...)

jamesc-grafana:
> Do it (maybe Monday morning at this point...)

Oh yeah, not doing a Friday afternoon deploys 😉

PR #988, opened 2025-05-19
feat: new get-latest-workflow-artifact action

zerok:
This allows the user to retrieve an artifact from the latest run of a specific workflow, making data exchange across workflows easier.


actions/get-latest-workflow-artifact/src/main.ts
```diff
@@ -0,0 +1,81 @@
...
+      throw new Error("GITHUB_WORKSPACE is not set");
+    }
+    path = workspace;
+  }
+
+  path = filepath.join(path, artifactName);
+
+  const client = github.getOctokit(ghToken);
+
+  const { data } = await client.rest.pulls.get({
```
dblinkhorn, line 28: Does it make sense to have some error handling in place for these requests and the others below (getting workflows, artifacts, etc)?

zerok: Not sure. It would probably make the output in the error case more "readable" but in the end the stack trace should be the most useful piece of info.

dblinkhorn (approved):

Issue #980, opened 2025-05-13
`login-to-gar`: Consider creating our own component for docker-credential-gcr binary


dsotirakis:
Consider moving away from `StevenACoffman/setup-docker-credential-gcr` so we can have full control and be secure over the action. The action lies in this repo: https://github.com/StevenACoffman/setup-docker-credential-gcr, and what it does looks simple to mimic.

We can either create a separate reusable workflow for it, or simply for now add it as a step in `login-to-gar` action. (we can start with the latter for now and convert it to a reusable workflow if needs be)

PR #996, opened 2025-05-26
fix: use custom step for docker-credential-gcr

dsotirakis:
Replacing the [upstream action](https://github.com/StevenACoffman/setup-docker-credential-gcr) so we can install the binary for different architectures, as well as removing sudo from the "untar" process. (will create a PR to the upstream repo, at least for the architectures - will leave the sudo handling to the author by creating an issue).

Note that we create a new directory under `/opt/` so it can be accessible from within a docker image, in case an action runs inside a container.

Part of: https://github.com/grafana/shared-workflows/issues/980

### DIFF
--- a/actions/login-to-gar/action.yaml
+++ b/actions/login-to-gar/action.yaml
@@ -59,11 +59,53 @@ runs:
       with:
         project_id: "grafanalabs-workload-identity"
         workload_identity_provider: "projects/304398677251/locations/global/workloadIdentityPools/github/providers/github-provider"
-    - name: Setup docker-credential-gcr
-      if: ${{ steps.auth_with_service_account.outputs.access_token == '' }}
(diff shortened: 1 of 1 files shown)


guicaulada (approved):
I think this is great, considering [we were the only ones](https://github.com/search?q=%22uses%3A+StevenACoffman%2Fsetup-docker-credential-gcr%22&type=code) using `StevenACoffman/setup-docker-credential-gcr` in all of GitHub, even though we were using the SHA, it always felt like a security incident waiting to happen.

grafanalf:
I believe this is broken, because as I can see from the logs in https://github.com/grafana/deployment_tools/actions/runs/15297604351/job/43030333860?pr=269738, installing `docker-credentials-gcr` requires `sudo`.

PR #1001, opened 2025-05-28
fix(login-to-gar): replace hardcoded opt dir with runner temp env var

dsotirakis:
Looks like non-root users cannot store things to `opt/` directory. It'd be better if we stored the binary in the temp dir on the runner, using `${ RUNNER_TEMP }`.

**Note**: 
(also seen as a comment in the action)
We are adding a `zizmor: ignore[github-env]` since `docker-credential-gcr` binary should be in the PATH. We are only adding a path to a binary that was just downloaded, so the risk should be minimal. credHelpers helper in the /.docker/config.json must be on the PATH. Also, RUNNER_TEMP is controlled by GitHub Actions and not accessible to attackers.

-----

Working examples:
* [Check terraform files](https://github.com/grafana/deployment_tools/actions/runs/15299827911/job/43042468555)
* [Build and push projects in docker](https://github.com/grafana/deployment_tools/actions/runs/15299828033)


actions/login-to-gar/action.yaml
```diff
@@ -64,7 +64,7 @@ runs:
       shell: bash
       env:
         DOCKER_CREDENTIAL_GCR_VERSION: "v2.1.28"
-      run: |
+      run: | # zizmor: ignore[github-env]
```
iainlane, line 67: this is in here twice - are both needed?

iainlane (approved):
[Αλίμονο!](https://gist.github.com/iainlane/3ec64ad0a53d481885feaad0569a30de)

PR #1009, opened 2025-06-04
feat!: only allow direct workload identity federation in login-to-gar

dsotirakis:
Removes old deprecated authentication after all GAR ops have moved to Direct WIF.

Part of: https://github.com/issues/assigned?issue=grafana%7Cdeployment_tools%7C259653


zerok (approved):
This is actually a breaking change, right? If so, please update the title to `feat!: ...` before merging 🙂

dsotirakis:
good call @zerok, thanks!

PR #1022, opened 2025-06-11
feat(get-latest-workflow-artifact): consider in-progress runs

zerok:
In some situations a workflow might get triggered that depends on an artifact while the producing workflow is still in-progress.

There might even be the case that the artifact hasn't been uploaded yet and so this scenario also allows for 5 retries taking place.

The action now also returns the status of the used workflow as output.

### DIFF
--- a/actions/get-latest-workflow-artifact/src/main.ts
+++ b/actions/get-latest-workflow-artifact/src/main.ts
@@ -11,6 +11,7 @@ async function main() {
     .split("/");
   const prNumber = parseInt(core.getInput("pr-number", { required: true }), 10);
   const artifactName = core.getInput("artifact-name", { required: true });
+  const considerInProgress = core.getInput("consider-inprogress") === "true";
   let path = core.getInput("path", { required: false });
 
   if (!path) {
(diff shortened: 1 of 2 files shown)


actions/get-latest-workflow-artifact/src/main.ts
```diff
@@ -32,50 +33,96 @@ async function main() {
...
+    );
+    if (artifact) {
+      console.log(`Found ${status} artifact`);
+      return { artifact, run };
+    }
+    console.log("No artifact found, retrying in 10s");
+    await new Promise((resolve) => {
+      setTimeout(() => resolve(), 10000);
+    });
+  }
```
iainlane, line 126:
```suggestion
  }
  
  return null;
```

actions/get-latest-workflow-artifact/src/main.ts
```diff
@@ -32,50 +33,96 @@ async function main() {
...
 
+async function getLatestArtifact(
+  client,
+  repoOwner,
+  repoName,
+  workflowId,
+  headSha: string,
+  status: string,
+  artifactName: string,
+): Promise<{ artifact: any; run: any } | null> {
```
iainlane, line 94: These should be given better types!

If you add this at the top

```typescript
import type { components } from "@octokit/openapi-types";

type GetArtifactResponse = components["schemas"]["artifact"];
type GHClient = ReturnType<typeof github.getOctokit>;
type WorkflowJobStatus = components["schemas"]["job"]["status"];
type WorkflowRun = components["schemas"]["workflow-run"];
```

this could become

```suggestion
async function getLatestArtifact(
  client: GHClient,
  repoOwner: string,
  repoName: string,
  workflowId: string,
  headSha: string,
  status: WorkflowJobStatus,
  artifactName: string,
): Promise<{ artifact: GetArtifactResponse; run: WorkflowRun } | null> {
```

actions/get-latest-workflow-artifact/src/main.ts
```diff
@@ -32,50 +33,96 @@ async function main() {
...
-    workflow_id: workflowId,
-    status: "completed",
-  });
 
-  if (runs.workflow_runs.length <= 0) {
-    throw new Error(`No workflow runs found for sha ${headSha}`);
-  }
+  // Now let's get all workflows associated with this sha:
+  // We start with in-progress ones if those are to be considered.
+  let found = null;
```
iainlane, line 39: needs to be typed, otherwise the assignments further down aren't valid, since `found` will have type `null`

```suggestion
  let found: { artifact: GetArtifactResponse; run: WorkflowRun } | null = null;
```

iainlane (commented):
I started reviewing this earlier but only just got around to finishing the review too late - sorry.

Hope these comments can be helpful for a followup..!

PR #1023, opened 2025-06-11
feat(login-to-gar): store credentials in temporary location by default

iainlane:
The `google-github-actions/auth` action which we use to authenticate to GCP for pushing to our artifact registries [creates its credentials file in the workspace][workspace_credentials].

It's a choice made by upstream to have subsequent Docker actions able to work transparently. But it comes with a big tradeoff. Active credentials are placed in the working directory. If workflow authors aren't aware of this, it's very easy to accidentally print or commit these. Consider a workflow which does some work and commits any resulting diff. If someone comes along later and adds an `auth` step, it's very unlikely that the author or reviewer will realise that the credentials file is about to be committed.

We've recognised this already by providing a `delete_credentials_file` option. By default we will configure Docker and then delete the credentials file. This is fine in most cases where folks just want to use `docker`, but it might not be suitable for other ways of accessing artifact registries such as `gcloud` or language SDKs which may not use credential helpers.

To make that last point fully true, and since we're now keeping the file around, we don't need to use `docker-credential-gcr` to configure Docker. We can use `gcloud`, which should be preinstalled in runner images in most circumstances, but we add `setup-gcloud` to be sure it's there. This should allow `docker`, `gcloud` and other clients to all work fine.

[workspace_credentials]: https://github.com/google-github-actions/auth/blob/0920706a19e9d22c3d0da43d1db5939c6ad837a8/src/main.ts#L144-L148


dsotirakis (approved):
sorry it took me so long - tried on my private repo and works great.
together with `tree` I tried `git status` and the json is there when `workspace_credentials: true` and it's not there when it's false (default)

looks great 💪

PR #1026, opened 2025-06-12
fix(get-latest-workflow-artifact): suggested fixes

zerok:
... from https://github.com/grafana/shared-workflows/pull/1022


iainlane (approved):
thanks for doing this!

PR #1099, opened 2025-07-03
feat(get-latest-workflow-artifact): consider runs triggered by comments

zerok:
If a workflow is triggered by a comment, we have no clear way to associate it with a PR through the head_sha. Instead we have to look at the time and title of the run. This is not completely reliable but should be "good enough" for most cases. Due to that it is disabled by default.


dsotirakis:
> The problem is: issue_comment does not reference the PR's branch as head but the target branch which defeats that solution. This is also the reason why this implementation tries to match workflow runs using PR title + time window (around when a comment was made).

I see now. I wonder if this is the expected behavior or if we should create an issue with github about this?

dsotirakis (approved):

PR #1127, opened 2025-07-10
chore(trigger-argo-workflow): use cli/v3

zerok:
Besides the "normal" updates this also moves the "submit" command into a proper subcommand which is necessary because otherwise the extraArgs would interfere with the rest of the argument parsing. Despite that, this change should not
be breaking since the `submit` command was also necessary in the previous implementation.

### DIFF
--- a/actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
+++ b/actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"io"
 	"log/slog"
 	"maps"
 	"os"
@@ -24,6 +25,7 @@ const (
 	flagNamespace        = "namespace"
 	flagParameter        = "parameter"
 	flagRetries          = "retries"
+	flagPrintConfig      = "print-config"
 	flagWorkflowTemplate = "workflow-template"
 )
 
@@ -43,13 +45,17 @@ func parseLogLevel(level string) (slog.Level, error) {
 }
 
 func main() {
+	runMain(os.Args, os.Stdout, os.Stderr)
+}
+
+func runMain(args []string, writer io.Writer, errWriter io.Writer) {
 	// If we're on a terminal, we use tint, otherwise if we're on GitHub actions
 	// we use `willabites/actionslog` to log proper Actions messages, otherwise
 	// we use logfmt.
 	var lv slog.LevelVar
 
 	logger := slog.New(
-		slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
+		slog.NewTextHandler(errWriter, &slog.HandlerOptions{
 			Level: &lv,
 		}),
 	)
@@ -71,6 +77,8 @@ func main() {
 	}
 
 	app := cli.Command{}
+	app.Writer = writer
+	app.ErrWriter = errWriter
 	app.Name = "Runs the Argo CLI"
 
 	app.Action = func(ctx context.Context, c *cli.Command) error {
@@ -81,6 +89,8 @@ func main() {
 		{
 			Name:            "submit",
 			SkipFlagParsing: true,
+			Writer:          writer,
+			ErrWriter:       errWriter,
 			Action: func(ctx context.Context, c *cli.Command) error {
 				return run(ctx, c, &lv, logger, "submit")
 			},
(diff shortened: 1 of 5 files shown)


iainlane:
Since we broke it once, any chance we could add some integration tests?

zerok:
> Since we broke it once, any chance we could add some integration tests?

Yupp, makes sense. Will try to come up with something tomorrow

zerok:
https://github.com/grafana/shared-workflows/pull/1127/commits/5a24116a93af5c11c94a2bbfd25ba5e64d6c581c should cover the scenario we ran into yesterday 🙂

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
```diff
@@ -69,35 +76,64 @@ func main() {
...
-			Usage:   "If true, the `--labels` argument will be added with values from the environment. This is forced for the `submit` command",
+			Name:     flagPrintConfig,
+			Required: false,
+			Usage:    "If set thie command will only print the gathered configuration and exist",
+		},
+		&cli.BoolFlag{
+			Name: flagAddCILabels,
+			Sources: cli.NewValueSourceChain(
+				cli.EnvVar("ADD_CI_LABELS"),
+			),
```
iainlane, line 110: q: Over on [the migration guide](https://cli.urfave.org/migrate-v2-to-v3/#__tabbed_9_2), they suggest a shorter form. Would that work?

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
```diff
@@ -173,7 +214,7 @@ func run(c *cli.Context, level *slog.LevelVar, logger *slog.Logger) error {
 		extraArgs = append(extraArgs, "--parameter", param)
 	}
 
-	extraArgs = append(extraArgs, c.Args().Tail()...)
+	extraArgs = append(extraArgs, c.Args().Slice()...)
```
iainlane, line 217: This one caught my eye. Looks like `Tail()` is still available, so is this a fix for a bug?

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+			},
+		},
+	}
+	for testName, test := range tests {
+		t.Run(testName, func(t *testing.T) {
+			for envName, envValue := range test.env {
+				t.Setenv(envName, envValue)
+			}
+			writer := &bytes.Buffer{}
+			runMain(test.args, writer, os.Stderr)
```
iainlane, line 59: just a nit: we could send the logs to `t.Log()` too with something like this:

```go
type testingWriter struct {
	t *testing.T
}

func (tw *testingWriter) Write(p []byte) (n int, err error) {
	scanner := bufio.NewScanner(bytes.NewReader(p))
	for scanner.Scan() {
		tw.t.Log(scanner.Text())
	}

	if err := scanner.Err(); err != nil {
		return 0, err
	}

	return len(p), nil
}
```

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+			if test.check != nil {
```
iainlane, line 60: `check` is never `nil` so we could remove this

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
zerok: Yes, this is a result of the `submit` command being an actual command.

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main.go
zerok: That would also work. Personally, I think the longer form is just more readable as it sets `Sources` to something that hints at a list of sources 🙂

actions/trigger-argo-workflow/cmd/trigger-argo-workflow/main_test.go
iainlane: what do you mean? that's not out!

iainlane (approved):

PR #1164, opened 2025-07-22
docs: add instructions for using Renovate

iainlane:
Our versioning schema doesn't work with Renovate out of the box. Fortunately it can be configured to handle it fine. Here we add instructions for users so that they can set their Renovate configuration up correctly.


zerok (approved):

guicaulada (dismissed):

iainlane:
FYI this seems to not be _quite_ correct. @dsotirakis is working on a fixed version so I'll assign this to him 🙂

PR #1348, opened 2025-09-24
feat: docker multi-arch reusable workflow

ricky-undeadcoders:
This pull request introduces a reusable workflow that uses a matrix strategy to perform native multi-arch docker builds for `arm64` and `amd64` builds. 

It uses the 3 new composite actions from https://github.com/grafana/shared-workflows/pull/1347.

### DIFF
--- a/.github/workflows/docker-build-push-multiarch.md
+++ b/.github/workflows/docker-build-push-multiarch.md
@@ -1,0 +1,86 @@
+# docker-build-push-multiaarch
+
+This is a reusable workflow, that uses Grafana's hosted runner to natively build and push multi-architecture docker
+images.
+
+This is effectively a wrapper around [docker-build-push-image](), [docker-export-digest](),
+and [docker-import-digests-push-manifest](), with a [matrix strategy]()k to determine which instance types to launch.
+
+# TODO: do we need QEMU?
+
+<!-- x-release-please-start-version -->
+
+```yaml
+name: Build a Docker Image
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  build-push-image:
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: grafana/shared-workflows/actions/docker-build-push-image@main # TODO: Fix version once released
+        with:
+          platforms: linux/arm64,linux/amd64
+          tags: |
+            ${{ github.sha }}
+            main
+          push: true
+          registries: "gar,dockerhub"
+```
+
+<!-- x-release-please-end-version -->
+
+## Inputs
+
+| Name                      | Type    | Description                                                                                                                                                                                                            |
+|---------------------------|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `build-args`              | String  | List of arguments necessary for the Docker image to be built.                                                                                                                                                          |
+| `build-contexts`          | String  | List of additional build contexts (e.g., name=path)                                                                                                                                                                    |
+| `buildkitd-config`        | String  | Configum buildkitium descriptium                                                                                                                                                                                       |
+| `buildkitd-config-inline` | String  | Inliumium configutorium buildkitium descriptium                                                                                                                                                                        |
+| `cache-from`              | String  | Where cache should be fetched from                                                                                                                                                                                     |
+| `cache-to`                | String  | Where cache should be stored to                                                                                                                                                                                        |
+| `context`                 | String  | Path to the Docker build context.                                                                                                                                                                                      |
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build-push-multiarch.yml
```diff
@@ -0,0 +1,344 @@
...
+        if: ${{ inputs.pre-build-script != '' }}
+        run: |
+          set -euo pipefail
+          echo "Running $PRE_BUILD_SCRIPT_PATH"
+
+          if [ -f "$PRE_BUILD_SCRIPT_PATH" ]; then
+            chmod +x "$PRE_BUILD_SCRIPT_PATH"
+            "$PRE_BUILD_SCRIPT_PATH"
+          else
+            echo "No script found at $PRE_BUILD_SCRIPT_PATH, skipping."
```
zerok, line 275: Is skipping enough here or should this actually be an error? It does hint a mistake in the configuration.

.github/workflows/docker-build-push-multiarch.md
```diff
@@ -0,0 +1,98 @@
...
+And supports building the following image types:
+
+- linux/arm64
+- linux/amd64
+
+## How it works
+
+This generates a matrix based off of the `platforms` input, then creates a job per platform that runs the composite
+actions [docker-build-push-image] and [docker-export-digest] to build and push docker images, and capture their digests.
+There is a then a final job that runs the composite action [docker-import-digests-push-manifest] to push the final
```
dsotirakis, line 20:
```suggestion
There is then a final job that runs the composite action [docker-import-digests-push-manifest] to push the final
```

.github/workflows/docker-build-push-multiarch.md
```diff
@@ -0,0 +1,98 @@
...
+<!-- x-release-please-start-version -->
+
+```yaml
+name: Build and Push and Push MultiArch
+
+on: push
+
+jobs:
+  build-push-multiarch:
+    uses: grafana/shared-workflows/.github/workflows/build-and-push-docker-multiarch.yml@rwhitaker/multi-arch-builds # TODO: Pin to version
```
dsotirakis, line 37: Keep in mind that you cannot really "release" reusable workflows - see [issue here](https://github.com/googleapis/release-please/issues/1366). Users will still be able to use the reusable workflow based on a commit sha, but release-please isn't able to create semver compatible versions. At least not yet!

.github/workflows/docker-build-push-multiarch.yml
```diff
@@ -0,0 +1,368 @@
+name: Push to artifact registry MultiArch
+# description: Resuable workflow to push to Google Artifact Registry using runners of different architectures.
```
dsotirakis, line 2: Do we want to uncomment that maybe?

.github/workflows/docker-build-push-multiarch.yml
```diff
@@ -0,0 +1,367 @@
...
+
+          if [ -f "$PRE_BUILD_SCRIPT_PATH" ]; then
+            chmod +x "$PRE_BUILD_SCRIPT_PATH"
+            "$PRE_BUILD_SCRIPT_PATH"
+          else
+            echo "::error::No script found at $PRE_BUILD_SCRIPT_PATH."
+            exit 1
+          fi
+        env:
+          PRE_BUILD_SCRIPT_PATH: ${{ inputs.pre-build-script }}
```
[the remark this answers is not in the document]
ricky-undeadcoders: Since this is being passed as an input... what would that look like here? Would we need the users to pass their script inline, or could we take the script file and parse it out and execute it somehow?

.github/workflows/docker-build-push-multiarch.yml
```diff
@@ -0,0 +1,367 @@
...
+          PLATFORM=$(echo "$ARCH_TO_PLATFORM_MAP" | jq -r --arg arch "${ARCH}" '.[$arch]')
+          echo "PLATFORM=${PLATFORM}" | tee -a "${GITHUB_ENV}"
+          echo "PLATFORM_PAIR=${PLATFORM//\//-}" | tee -a "${GITHUB_ENV}"
+        env:
+          ARCH: ${{ matrix.arch }}
+      - name: Checkout code
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          persist-credentials: false
+      - name: Run custom build script if it exists
```
iainlane: Just a bit of feedback -- take it or leave it 🙂

I would consider leaving the scripting components out of the "boxed up" reusable workflow. I think it's OK enough to say that if you need to run code before/after your build then you can use the "parts-based" workflow. It gives people the flexibility to do more than you can do with a script too, for example to use third party actions.

PR #1381, opened 2025-10-06
feat(with-post-step): create with-post-step action

eloymg:
Create with-post-step action using as a source this code: https://github.com/pyTooling/Actions/tree/main/with-post-step.

context why it's necessary: https://github.com/actions/runner/issues/1478


zerok (commented):
Do you already have a ue-case for this in mind? Are there any particular reasons why we don't simply use the upstream action?

Issue #137, opened 2024-05-31
Generate action READMEs from the YAML


jdbaldry:
I don't trust that any of the READMEs' inputs and outputs match the actual action specification.

I think a quick tool that generates the READMEs from a template that can be used for all actions would be better than relying on maintainers and contributors to keep the two in sync. Of course, CI should enforce that the README matches the action. We could even add an additional workflow that regenerates the READMEs on a branch so that it doesn't need to be performed locally.

I understand that READMEs help with discoverability so I'm not sure that it's justified to remove them in the mean time.

ricky-undeadcoders:
I have 2 PRs open that should complete this task.

**The first (#1454) is a new script** at `scripts/generate-input-output-docs/`. It parses a composite action or a reusable workflow, pulls out every input and output, and renders them as a markdown table. It's the draft script from this issue, finished off: it now walks the whole repo rather than taking one file at a time, writes the tables into the docs itself, and has tests.

It has four subcommands:

```sh
cd scripts/generate-input-output-docs

go run . generate -root-dir ../../   # rewrite the tables in every doc in the repo
go run . check    -root-dir ../../   # report which docs are stale, without writing
go run . parity   -root-dir ../../   # check a workflow still matches the action it wraps
go run . print -file ../../actions/aws-auth/action.yaml   # dump one file's tables to stdout
```

That PR deliberately changes no action READMEs. Every action here is its own release-please package, so a PR that touches 37 of them would bump 37 versions. Keeping the tool on its own means it lands under `scripts/`, which isn't a package, and triggers no releases.

**The second (#2276) implements the CI check and updates all the docs.** It regenerates every action README and reusable workflow doc from its YAML, and adds `.github/workflows/check-action-docs.yaml`, which regenerates in CI and fails if the committed docs don't match. It's stacked on #1454, so that one merges first.

Between them they cover all three checks from the issue:

---

### 1. `docker-build-push-multiarch.yml` vs `docker-build-push-image/action.yaml`

Done, as `go run . parity`. The rules live in `scripts/generate-input-output-docs/parity.yaml` and run in CI.

It compares the input names and output names on both sides in both directions. Anything on one side but not the other is a failure, unless it's listed in the rule as deliberately one-sided with a reason — the runner-type inputs, for instance, configure the workflow's fan-out and have no meaning for the action.

[...]

PR #1454, opened 2025-10-24
feat: add action and workflow input/output docs generator

ricky-undeadcoders:
Adds a tool that parses composite actions and reusable workflows and renders their inputs and outputs as markdown tables, so the docs can be generated from the YAML instead of maintained by hand.

This is the generator only. **The regenerated docs and the CI enforcement job are in the stacked PR** — this one deliberately changes no action READMEs, so it touches no release-please package and triggers no version bumps.

Picks up the draft script from a year ago and finishes it.

## What it does

```sh
cd scripts/generate-input-output-docs

go run . generate -root-dir ../../   # rewrite every doc in the repo
go run . check    -root-dir ../../   # report drift without writing
go run . parity   -root-dir ../../   # workflow vs the action it wraps
go run . print -file ../../actions/aws-auth/action.yaml   # one file, to stdout
```

Discovers composite actions at `actions/*/action.y{a,}ml` (the repo uses both spellings) and reusable workflows in `.github/workflows` that declare `on.workflow_call`, documenting them in `README.md` and a sibling `.md` respectively.

## Design notes

**Marker-based injection.** Tables are written between `<!-- BEGIN_INPUTS -->` / `<!-- END_INPUTS -->`. Only the text between markers is replaced, so hand-written prose, footnotes and markdownlint pragmas survive regeneration. Generating whole READMEs from a template would have flattened all of that.

**Required and Default come from the YAML.** That is the point — a default documented in a sentence goes stale, a generated one cannot. Outputs render as Name/Description only, since neither schema gives an output a type, default or required flag.

**Prettier is a hard dependency, pinned.** The tool shells out to prettier and requires the exact version in `package.json`. Prettier is the repo's formatter of record and it rewrites table *content*, not just padding — 3.8.3 renders an asterisk in a cell as `\*.out` and 3.9.6 renders it as `*.out`. Generating with a different version than CI produces a one-byte difference and a drift failure on a PR whose docs were just regenerated. Requiring the pin makes local output identical to CI by construction, so `bun install` is needed before `generate`.

**Parity checking** (`parity.yaml`) asserts that `docker-build-push-multiarch` still exposes the same inputs and outputs as the `docker-build-push-image` action it wraps. This is what the original draft of the script was written for. One-sided names are listed with a reason, and a listed name that no longer exists is itself reported, so the exceptions cannot quietly rot.

It found real drift on its first run: the `permanent-inputs.yml` list sketched a year ago (`server-size`, `pre-build-script`, `post-build-script`) was entirely stale — none of those inputs exist on the workflow any more. That file is deleted here. It also found three genuine forwarding gaps (`annotations`, `builder`, `metadatajson`), left as documented TODOs since changing a released workflow's interface deserves its own PR.

## Known limitation

Composite actions have no `type` field in GitHub's schema, so the type is inferred from the default value. **A boolean input with no default is reported as `String`** — nothing in the YAML distinguishes it from a string. This mislabels exactly one input across all 191 in the repo (`docker-build-push-image.push`); giving it an explicit `default: "false"` would fix both the type and the docs, but that is a behaviour change and belongs elsewhere.

## Tests

66 tests. Three need the pinned prettier and skip without it, so `go test` works before `bun install`; CI installs it so they never silently skip. Coverage includes marker injection (replace, bootstrap, append, prose preservation, idempotence), malformed-marker rejection, the shorthand `on: push` / `on: [push]` forms, pipe and angle-bracket escaping, code-span fence widening, and the parity allow-list logic including stale entries.

Part of #1466. Groundwork for #137.


dblinkhorn:
@ricky-undeadcoders is this change still needed?

scripts/generate-input-output-docs/main.go
```diff
@@ -0,0 +1,256 @@
...
+Commands:
+  generate   Rewrite the generated tables in every doc file in the repo
+  check      Report which doc files are out of date, without writing (exit 1 on drift)
+  parity     Verify a reusable workflow forwards the inputs/outputs of its action
+  print      Print the tables for a single YAML file to stdout
+
+Run "generate-input-output-docs <command> -h" for the flags of a command.
+`
+
+func run(args []string) error {
```
zerok: nit: Just curious but why didn't you use something like urfave/cli/v3 here? Would make the parsing and subcommands much easier :-)

scripts/generate-input-output-docs/format.go
```diff
@@ -0,0 +1,159 @@
...
+// The repo-local install is preferred because it is the pinned version by
+// definition; bunx and npx resolve it too once dependencies are installed. A
+// global prettier comes last and is only accepted if its version happens to
+// match, since a contributor's global install is the most likely source of skew.
+func candidateFormatters(root string) [][]string {
+	return [][]string{
+		{filepath.Join(root, "node_modules", ".bin", "prettier")},
+		{"bunx", "prettier"},
+		{"npx", "--no-install", "prettier"},
+		{"prettier"},
```
zerok: Since the repo is pretty much bun-based, do we need these two options?

scripts/generate-input-output-docs/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/grafana/shared-workflows/scripts/generate-input-output-docs
+
+go 1.24.2
+
+require gopkg.in/yaml.v3 v3.0.1
```
zerok: For a follow-up: Please migrate to go.yaml.in/yaml/v3 or even newer. Should be pretty much a drop-in replacement.

scripts/generate-input-output-docs/inject.go
```diff
@@ -0,0 +1,213 @@
...
+func (s Section) Heading() string {
+	switch s {
+	case SectionOutputs:
+		return "## Outputs"
+	default:
+		return "## Inputs"
+	}
+}
+
+func (s Section) beginMarker() string { return fmt.Sprintf("<!-- BEGIN_%s -->", s) }
```
zerok, line 26: The PR description states that only BEGIN_INPUTS/END_INPUTS are handled 🙂

scripts/generate-input-output-docs/inject.go
```diff
@@ -0,0 +1,213 @@
...
+func (s Section) endMarker() string   { return fmt.Sprintf("<!-- END_%s -->", s) }
+
+// emptyPlaceholder stands in when a section's markers exist but the action or
+// workflow declares nothing, so the block is never left as blank markdown.
+const emptyPlaceholder = "_None._"
```
zerok, line 31: This does not have a test-case

scripts/generate-input-output-docs/inject.go
```diff
@@ -0,0 +1,213 @@
...
+	return appendSection(doc, section, block), nil
+}
+
+// validateMarkers rejects marker arrangements that cannot be updated safely.
+//
+// Each of these is a mistake rather than a state worth tolerating: left alone
+// they produce a doc that is stable on regeneration, so the CI drift check stays
+// green while the content silently rots.
+func validateMarkers(doc string, section Section) error {
+	begins := strings.Count(doc, section.beginMarker())
```
zerok: Perhaps as a follow-up but would it perhaps make sense to restrict those markers to be the only items on a given line and with no whitespace before and after?

scripts/generate-input-output-docs/parity.go
```diff
@@ -0,0 +1,160 @@
...
+		field, quoteList(stale), path,
+	)}
+}
+
+// joinRoot resolves a rule's repo-relative path against the configured root.
+func joinRoot(root, path string) string {
+	return filepath.Join(root, filepath.FromSlash(path))
+}
+
+func toSet(names []string) map[string]bool {
```
zerok: nit: use `map[string]struct{}` for simulating sets.

scripts/generate-input-output-docs/render.go
```diff
@@ -0,0 +1,181 @@
...
+// `<base-ref>..<commit-sha>`, `gh project view <number> --owner <org>`. Markdown
+// treats those as raw HTML tags, so the renderer swallows them and the reader
+// sees `:@` or `..`. Escaping makes them display literally.
+//
+// Text inside a code span is skipped: backticks already render their contents
+// literally, and an `&lt;` written there would show up as the entity itself
+// rather than as `<`.
+func escapeAngleBrackets(s string) string {
+	var b strings.Builder
+	b.Grow(len(s))
```
zerok, line 133: q: is this an optimization for the case where no `<` is encountered?

zerok (dismissed):
Looks good to me! Especially the table generation logic looks scary, though ^_^ I have a couple of questions but none of them blocking.

scripts/generate-input-output-docs/main.go
ricky-undeadcoders: So, originally I did, but I removed it because generate-catalog-info uses stdlib flag and has no CLI dependency, so this matches it, and it's one fewer dep for Renovate. But you're right, urfave would be less code. If there's more coding to do once I'm through reviewing the following comments then I'll add it in.

scripts/generate-input-output-docs/format.go
ricky-undeadcoders: Ah, good catch. We don't. I'll pull them out.

scripts/generate-input-output-docs/main.go
ricky-undeadcoders: Done — switched to urfave/cli/v3. You were right, it's cleaner: subcommand routing, required-flag enforcement and help text all come free, and it let me delete a bunch of hand coded nonsense.

Two notes for anyone reading later: it accepts both -flag and --flag, so the existing CI invocations kept working unchanged, and --root-dir now defaults to ../.. on every subcommand since the tool is always run from its own directory. go.mod is still only two direct requires — urfave/cli/v3 pulls no transitive dependencies.

scripts/generate-input-output-docs/format.go
ricky-undeadcoders: Done — dropped both.

scripts/generate-input-output-docs/go.mod
ricky-undeadcoders: Done in this PR rather than as a follow-up, since it was a one-line swap plus go mod tidy. Now on go.yaml.in/yaml/v3 v3.0.5.

scripts/generate-input-output-docs/inject.go
ricky-undeadcoders: Fixed

scripts/generate-input-output-docs/inject.go
ricky-undeadcoders: Good catch — and it sent me looking at coverage properly, which turned up a much worse gap.

emptyPlaceholder itself was already covered: TestInjectEmptySectionWithMarkersGetsPlaceholder passes an empty table with the markers present and asserts the placeholder replaces the stale content, with TestInjectSkipsEmptySectionWithNoMarkers covering the inverse.

But the neighbouring helpers in that hunk — Heading(), beginMarker() and endMarker() — only had indirect coverage through Inject. That's a fair gap: they're derived from the Section constant via fmt.Sprintf, so a typo would silently rename the markers in every generated doc at once, and the Inject tests wouldn't catch it because they build their expectations from the same helpers and would be self-consistently wrong. Added TestSectionHelpers to pin the exact output of all three for both sections.

While measuring I another hole: runGenerate was at 0% coverage. It took *cli.Command, so the loop that discovers targets, reads each doc and writes the result had no tests — including the branch that creates a doc from scratch for an action with no README, which is precisely the case the CI drift check exists to catch and the reason that check needed git add -A.

So I've extracted it as GenerateDocs(GenerateOptions), free of CLI types, with runGenerate as a thin adapter over it. Six tests now run it against a throwaway repository (one composite action, one reusable workflow, one ordinary on: push workflow): doc creation, ordinary workflows being skipped, no-op on a second run, check-only reporting drift without touching disk, and prose outside the markers surviving regeneration.

Package coverage went 72.2% → 77.2%, and Target.Name() / rel() went from 0% to covered. That also flushed out three fields that were written or parsed and never read — Spec.Kind, Spec.Path and rawIO.Value — now removed.

scripts/generate-input-output-docs/inject.go
ricky-undeadcoders: Done

scripts/generate-input-output-docs/parity.go
ricky-undeadcoders: Done

scripts/generate-input-output-docs/render.go
ricky-undeadcoders: Not that case specifically — b.Grow(len(s)) preallocates the builder to the input length. Escaping only ever grows the string (< → &lt;), so len(s) is a guaranteed lower bound on the output and this avoids the early reallocations. When there's no < at all it happens to be exactly right-sized, which is the common case.

zerok (approved):
Thank you ❤️

PR #1471, opened 2025-10-31
feat(build-push-to-dockerhub): add labels input

martincostello:
Allow custom labels to be added to container images.

### DIFF
--- a/actions/build-push-to-dockerhub/action.yaml
+++ b/actions/build-push-to-dockerhub/action.yaml
@@ -11,7 +11,8 @@ inputs:
     required: true
   labels:
     description: |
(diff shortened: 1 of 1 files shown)


martincostello:
I haven't investigated migrating fully yet, but we only need pushing to Docker Hub and haven't investigated/tested other CPU architectures for the image yet. For now this is just to allow us to resolve https://github.com/grafana/docker-otel-lgtm/issues/837.

zerok:
Okidoki 🙂  Since both fields have the same name, I think it would be fine to still have that in the "old" action 😄

zerok (approved):

PR #1568, opened 2025-11-24
feat: support passing 'builder' parameter in push-to-gar-docker

sd2k:
We are trying to use the [buildkit-cache-dance][1] method to get better caching of our Go build cache in CI, but that requires having control over the buildx builder in our pipelines. This commit adds the 'builder' input to the push-to-gar-docker shared action, and conditionally creates a new builder if it isn't provided.

[1]: https://github.com/reproducible-containers/buildkit-cache-dance/


sd2k:
Yep! we're using it [here](https://github.com/grafana/grafana-assistant-app/actions/runs/19748771004/job/56588105427?pr=3477).

README updated :slightly_smiling_face:

PR #1603, opened 2025-12-09
feat: shared workflow for sbom export from socket.dev API

ezebunandu:
Adds a shared workflow that can be used to export the SPDX SBOM for a repo enrolled in the socket.dev GitHub app based on the most recent scan.

### DIFF
--- a/actions/socket-export-sbom/action.yml
+++ b/actions/socket-export-sbom/action.yml
@@ -52,5 +52,5 @@ runs:
         if [[ -n "$OUTPUT_FILE" ]]; then
           OUTPUT_FILE=$(basename "$OUTPUT_FILE")
         fi
-        go run $ACTION_PATH/main.go $REPO_NAME $OUTPUT_FILE
+        go run $ACTION_PATH/main.go $REPO_NAME "${ACTION_PATH}/${OUTPUT_FILE}"
         echo "path=$ACTION_PATH/$OUTPUT_FILE" >> $GITHUB_OUTPUT


actions/socket-export-sbom/main.go
```diff
@@ -0,0 +1,132 @@
...
+
+	data, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, resp.StatusCode, err
+	}
+	return data, resp.StatusCode, nil
+}
+
+func (client *Client) GetRepoLastScanID(name string) (string, error) {
+	path := fmt.Sprintf("/orgs/grafana/repos/%s", name)
```
guicaulada: Same as `ExportSBOM`, I think it would be best if we could parametrize the org.
```suggestion
func (client *Client) GetRepoLastScanID(org, name string) (string, error) {
	path := fmt.Sprintf("/orgs/%s/repos/%s", org, name)
```

guicaulada (commented):
I left a few comments, they are mostly `nits`, there's no blocker, they are only recommendations. 🙂 

The only blocking thing is that this should be added to our `release-please-config.json` similar to [this diff](https://github.com/grafana/shared-workflows/commit/eed69781bdf177e1c126172712690dae79b0bc11#diff-c55d4dcb68c348b2c06d263819702fbce72eeeb35b662956d02049a5a18fcf2b).

actions/socket-export-sbom/action.yml
```diff
@@ -0,0 +1,45 @@
...
+    default: "grafana"
+  output_file:
+    description: "Output file path for the SBOM"
+    required: false
+
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Go
+      uses: actions/setup-go@v5
```
guicaulada: Pin this action to a hash.
```suggestion
      uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
```

actions/socket-export-sbom/README.md
```diff
@@ -0,0 +1,63 @@
...
+      - name: Get vault secrets
+        id: vault-secrets
+        uses: grafana/shared-workflows/actions/get-vault-secrets@get-vault-secrets/v1.3.0
+        with:
+          repo_secrets: |
+            SOCKET_API_TOKEN=socket:SOCKET_SBOM_API_KEY
+          export_env: false
+
+      - name: Export SBOM from Socket
+        uses: ./socket-sbom
```
guicaulada: This should point to the action in shared workflows:
```suggestion
        uses: grafana/shared-workflows/actions/socket-export-sbom@socket-export-sbom/v0.1.0
```

actions/socket-export-sbom/README.md
```diff
@@ -0,0 +1,63 @@
...
+        with:
+          socket_api_token: ${{ fromJSON(steps.vault-secrets.outputs.secrets).SOCKET_API_TOKEN }}
+          output_file: ${{ inputs.output_file }}
+
+      - name: Upload SBOM artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: "sbom"
+          path: socket-sbom/${{ inputs.output_file }}
```
guicaulada: The `socket-sbom` folder seems like something that only makes sense if this action is in the same repository as it's being called, when using it as a composite action, the output file will be in `${{ github.action_path }}` which can't be used from the calling workflow. I would use an output here, such as `${{ steps.export-sbom.outputs.path }}`
```suggestion
          path: ${{ steps.export-sbom.outputs.path }}
```

But the ID would have to be defined on the `Export SBOM from Socket`.

kelnage:
> I wouldn't be able to call from this repo until it's merged or is that possible?

Yup, you can call any version of an Action in GitHub regardless of whether it exists in the default branch, you just point it to the appropriate hash. E.g., you should be able to use:

```
uses: grafana/shared-workflows/actions/socket-export-sbom@b1046c7e43ec4525a60fbc785cbb607801366597
```

PR #1661, opened 2026-01-17
refactor: socket sbom export

ezebunandu:
refactors the socket-export-sbom action to be package-oriented, with tests.


actions/socket-export-sbom/socket.go
```diff
@@ -0,0 +1,144 @@
...
+	HTTPClient *http.Client
+}
+
+func NewClient(apiKey, org string) *Client {
+	return &Client{
+		APIKey:  apiKey,
+		BaseURL: BASEURL,
+		Org:     org,
+		HTTPClient: &http.Client{
+			Timeout: 5 * time.Second,
```
guicaulada: A 5 seconds timeout might be insufficient for large SBOMs, wouldn't be better to use the default 30s or make it configurable?

actions/socket-export-sbom/socket_test.go
```diff
@@ -0,0 +1,121 @@
...
+	got, err := c.GetSBOM("loki")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !cmp.Equal(want, got) {
+		t.Error(cmp.Diff(want, got))
+	}
+}
+
+func getTestSBOM() socket.SBOM {
```
guicaulada: I think only happy paths are being tested. We should probably add tests for 404 responses, invalid JSON formats, network errors, etc.

actions/socket-export-sbom/socket.go
```diff
@@ -0,0 +1,144 @@
...
+	return ParseSBOM(data)
+}
+
+func (client *Client) makeAPIRequest(URI string) ([]byte, error) {
+	URL := fmt.Sprintf("%s%s", client.BaseURL, URI)
+	authHeaderValue := fmt.Sprintf("Bearer %s", client.APIKey)
+	req, err := http.NewRequestWithContext(context.Background(), "GET", URL, nil)
+	if err != nil {
+		return nil, nil
+	}
```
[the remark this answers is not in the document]
ezebunandu: Silly oversight there. Will fix.

actions/socket-export-sbom/socket.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+type Option func(*Client)
+
+func WithBaseURL(url string) Option {
+	return func(c *Client) {
+		if url == "" {
+			c.BaseURL = BASEURL
+		}
+		c.BaseURL = url
```
guicaulada: This doesn't look right `c.BaseURL = url` always executes, so even though when `url == ""` you set `c.BaseURL = BASEURL` it gets replaced by `c.BaseURL = url` which ends up setting it to empty string.

This should be:
```suggestion
		if url == "" {
			c.BaseURL = BASEURL
		} else {
			c.BaseURL = url
		}
```

Or to keep it simpler, set `BaseURL: BASEURL` on the `NewClient` function and use:
```suggestion
		if url != "" {
			c.BaseURL = url
		}
```

actions/socket-export-sbom/cmd/export-sbom/main.go
```diff
@@ -0,0 +1,48 @@
...
+    Usage: main.go <repo name> <output file>
+    `
+	key := os.Getenv("SOCKET_API_TOKEN")
+	if key == "" {
+		fmt.Fprintln(os.Stderr, "SOCKET_API_TOKEN not provided")
+		os.Exit(1)
+	}
+	org := os.Getenv("SOCKET_ORG")
+	if org == "" {
+		fmt.Fprintln(os.Stderr, "Please specify socket org name, e.g. 'grafana'")
```
guicaulada: nit: Most error messages are all lowercase following Go conventions, this is the only other one with capitalization.

actions/socket-export-sbom/socket_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/grafana/shared-workflows/actions/socket"
```
guicaulada: Shouldn't this be `socket-export-sbom`?
```suggestion
	"github.com/grafana/shared-workflows/actions/socket-export-sbom"
```

guicaulada (commented):
Thank you for working on all these suggestions!

These are the last 2 minor nits and a concern over the module name not matching the directory name.

After this should be good to go!

Sorry for multiple reviews couldn't catch everything on the first ones.

PR #1737, opened 2026-02-20
feat: add component digest and artifact handling actions

s-gohel:
### Summary

This PR introduces two new composite GitHub Actions that together implement selective deployment — only redeploying components whose images actually changed, while reusing previous digests for unchanged components.

These actions are designed to complement the existing `component-change-detection` action and are intended to be adopted by services migrating away from inline shell scripts in their build/deploy workflows.

---

#### `save-component-digest`

Called once per component in the **build job** (typically inside a matrix), immediately after a Docker push. It writes two files to disk:

- `component-digests/<component_name>.txt` — the full `<dockertag>@<digest>` string
- `component-digests/dockertag.txt` — the short commit SHA used as the Docker tag

The `_digest` suffix is automatically stripped from the component name when forming the filename (e.g. `grafana_com_api_digest` → `grafana_com_api.txt`). These files are uploaded as artifacts so the deploy job can consume them.

#### `component-selective-deploy`

Called in the **deploy job** after `component-change-detection` has run and all component digest artifacts have been downloaded. It runs four steps in sequence:

1. **Aggregate** — reads `component-digests/*.txt` and builds `new-digests.json`
2. **Select** — for each component, picks the new digest if changed, or the previous digest from `component-tags.json` if unchanged
3. **Validate** — asserts every selected digest is non-empty and matches `<tag>@sha256:<64-hex>` format; fails the job if any are malformed
4. **Update** — rewrites `component-tags.json` with the new commitSHA and digest for changed components only, leaving unchanged entries intact

Outputs `selected-digests-json` (a JSON object keyed as `<component>_digest`) and `dockertag` for use in downstream deployment steps.

---

### Testing

Both actions include CI workflows (`test-save-component-digest.yml`, `test-component-selective-deploy.yml`) that exercise realistic scenarios end-to-end on GitHub-hosted runners:

- **`save-component-digest`**: verifies correct file contents and that the `_digest` suffix is stripped from the filename
- **`component-selective-deploy`**: simulates a three-component build where two components changed and one did not; validates that changed components use new digests and unchanged components reuse old digests from `component-tags.json`


zerok:
Hm... do you already have a scenario where this is used? If this is for a single product and still in the testing phase, I think it would make more sense to put that implementation into a product-specific repository.

PR #1861, opened 2026-04-09
feat(zizmor): optional vendored-path excludes (security-appsec#326)

isaiah-grafana:
This adds optional **`.github/zizmor-collection-ignore`**: if it’s missing, zizmor still runs on `.` like before. If it’s there, we `find` workflow/action/dependabot paths with your prefixes pruned out (so vendored trees with nested workflows don’t get pulled in). Big lists get split into batches of 400 and SARIF is merged with `jq`.

Tracks https://github.com/grafana/security-appsec/issues/326

**After this merges:** pin `grafana/security-github-actions` `self-zizmor.yaml` to `grafana/shared-workflows/.github/workflows/reusable-zizmor.yml@<this merge SHA>`, then land the `deployment_tools` PR that adds the ignore file.

Details are in `.github/workflows/reusable-zizmor.md`.


.github/workflows/reusable-zizmor.yml
```diff
@@ -399,21 +453,83 @@ jobs:
...
-            --cache-dir "${ZIZMOR_CACHE_DIR}" \
-            ${ZIZMOR_CONFIG_PATH:+--config "${ZIZMOR_CONFIG_PATH}"} \
-            ${RUNNER_DEBUG:+"--verbose"} \
-            ${ZIZMOR_EXTRA_ARGS:+${ZIZMOR_EXTRA_ARGS}} \
-            . \
-            > results.sarif
-          exit_code=$?
-          # Only fail on exit code 1 (zizmor crash). 0 = no findings, 10-14 = findings by severity.
-          if [ "${exit_code}" -eq 1 ]; then exit 1; fi
+          set -euo pipefail
```
zerok, line 442: The complexity here is enormous. Would it perhaps make sense to move that into Python or JavaScript to make things like the path trimming etc. easier to read and test?

actions/zizmor-collection-paths/collection_paths.py
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env python3
+"""Build explicit zizmor scan paths when ignore prefixes are configured."""
+
+from __future__ import annotations
```
zerok: I'm not completely sure that you need this feature flag.

actions/zizmor-collection-paths/collection_paths.py
```diff
@@ -0,0 +1,141 @@
...
+    if not s or s.startswith("#"):
+        return None
+    if s.startswith("/"):
+        s = s[1:]
+    if s.endswith("**"):
+        s = s[:-2]
+    if s.endswith("/*"):
+        s = s[:-2]
+    while s.endswith("/"):
+        s = s[:-1]
```
zerok: You can probably use string.lstrip and string.rstrip for most of these instead of doing list-cutting 🙂

.github/workflows/reusable-zizmor.yml
```diff
@@ -399,21 +433,83 @@ jobs:
...
```
zerok, line 442: Would the whole batching also perhaps make sense as Python script?

actions/zizmor-collection-paths/collection_paths.py
isaiah-grafana: Yeah the naming is a little awkward here. `use_explicit_paths` sounds like a feature flag but it is really just wiring so repos without `.github/zizmor-collection-ignore` keep the same behavior as today where we run zizmor on `.`. If the ignore file exists and we parsed prefixes then we need explicit targets because running on `.` mostly defeats the point and you still end up pulling vendored nested workflow trees in through the repo root scan. If you want I can rename that output to something clearer like `explicit_targets` or `scan_mode`. Either way it is just a simple signal so the reusable workflow does not have to redo ignore parsing in bash or yaml.

actions/zizmor-collection-paths/collection_paths.py
isaiah-grafana: Yep I can clean that up.

.github/workflows/reusable-zizmor.yml
isaiah-grafana: Yeah I agree with you. I kept it in the workflow for now because it sits right next to the exact zizmor command line and writing `results.sarif`, and that bash section was getting reworked a lot while we were stabilizing the workflow checks. Moving the batching and SARIF merge into a small python helper next to `collection_paths.py` should be easier to test and less brittle imo. I can do that refactor in this PR.

actions/zizmor-collection-paths/collection_paths.py
zerok: I'm talking about the `from __future__ import annotations` line.

isaiah-grafana:
Opening a same-repo replacement PR so org rulesets/tests resolve against `grafana/shared-workflows` directly: https://github.com/grafana/shared-workflows/pull/1893\n\nClosing this fork-head PR to avoid confusion.

PR #1877, opened 2026-04-15
feat(zizmor): add Zizmor config validation

mikefat:
This PR adds a step in the reusable zizmor action that will pre-validate the zizmor config file to be used in the run. If validation fails, the job stops and zizmor is not executed.

The validation will fail on any file that does any of the following:
- Defines any of these audit blocks under `rules` (no `disable`, `ignore`, or `config` is allowed): `insecure-commands`, `template-injection`, `impostor-commit`, `known-vulnerable-actions`, `ref-confusion`.
- Sets `rules.unpinned-uses.disable`
- Sets `rules.unpinned-uses.config.policies` with a universal [`"*": any`](https://docs.zizmor.sh/audits/#unpinned-uses) entry (all `uses:` clauses may stay unpinned). Scoped policies such as `actions/*: any` or `grafana/*: any` remain valid.

Any other suggestions for things to be on the lookout/block are welcome

### DIFF
--- a/.github/workflows/reusable-zizmor.md
+++ b/.github/workflows/reusable-zizmor.md
@@ -157,6 +157,24 @@ uses: actions/checkout@v3 # zizmor: ignore[artipacked]
 
 [zizmor-ignore-comment]: https://woodruffw.github.io/zizmor/usage/#with-comments
 
+## Repo-local `zizmor.yml` policy gate
+
+When `always-use-default-config` is `false` and the calling repository uses a **repo-local** `zizmor.yml` or `.github/zizmor.yml` (so zizmor discovers that file instead of the Grafana default from `shared-workflows`), the workflow **validates that file before running zizmor**. If validation fails, the job stops and zizmor is not executed.
+
+The gate is **skipped** when:
+
+- `always-use-default-config` is `true` (only the Grafana default from this repository is used), or
+- There is no repo-local config file, or
+- The run uses the fetched default via `--config` (no repo-local file in play).
+
+The policy rejects configs that:
+
+- Define `rules.insecure-commands` or `rules.template-injection` (any keys under those audits),
+- Set `rules.unpinned-uses.disable`,
+- Set `rules.unpinned-uses.config.policies` with a universal [`"*": any`](https://docs.zizmor.sh/audits/#unpinned-uses) entry (all `uses:` clauses may stay unpinned). Scoped policies such as `actions/*: any` or `grafana/*: any` remain valid.
+
+Inline `# zizmor: ignore[...]` comments in workflow files are unchanged; this gate applies only to the repo-local YAML config file.
+
 ## Configuration
 
 zizmor [can be configured][zizmor-config] with a `zizmor.yml` or
--- a/.github/workflows/reusable-zizmor.md
+++ b/.github/workflows/reusable-zizmor.md
@@ -159,7 +159,7 @@ uses: actions/checkout@v3 # zizmor: ignore[artipacked]
 
 ## Repo-local `zizmor.yml` policy gate
 
-When `always-use-default-config` is `false` and the calling repository uses a **repo-local** `zizmor.yml` or `.github/zizmor.yml` (so zizmor discovers that file instead of the Grafana default from `shared-workflows`), the workflow **validates that file before running zizmor**. If validation fails, the job stops and zizmor is not executed.
+When `always-use-default-config` is `false` and the calling repository uses a **repo-local** `zizmor.yml` or `.github/zizmor.yml` (so zizmor discovers that file instead of the Grafana default from `shared-workflows`), the workflow **validates that file before running zizmor**. If validation fails, the job stops and zizmor is not executed. Validation uses the composite action [`actions/validate-zizmor-config`](../../actions/validate-zizmor-config) from this repository at the **same ref** as this reusable workflow (via `job-workflow-ref` outputs). **Set up Zizmor configuration** writes the path to validate as step output `repo-local-zizmor-config` when a repo-local file is used; that is separate from `zizmor-config`, which is only set when the run uses **`--config`** with the downloaded Grafana default.
 
 The gate is **skipped** when:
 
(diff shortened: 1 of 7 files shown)


mikefat:
Testing this changes in the private repo [here](https://github.com/grafana/test-new-zizmor-workflow/pull/1) and it seems to be working as expected 👍

.github/workflows/reusable-zizmor.yml
```diff
@@ -388,6 +390,188 @@ jobs:
...
+              parser = argparse.ArgumentParser(description=__doc__)
+              parser.add_argument("config_path", type=Path, help="Path to zizmor.yml or .github/zizmor.yml")
+              args = parser.parse_args()
+              path: Path = args.config_path
+
+              if not path.is_file():
+                  _fail(path, "config file does not exist or is not a file")
+
+              text = path.read_text(encoding="utf-8")
+              if len(text) > 256 * 1024:
```
zerok: Why that limitation? It looks a bit arbitrary.

.github/workflows/reusable-zizmor.yml
```diff
@@ -388,6 +390,188 @@ jobs:
...
+          CONFIG_TO_VALIDATE: ${{ steps.setup-config.outputs.repo_local_zizmor_config }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          SCRIPT="${RUNNER_TEMP}/validate_zizmor_config.py"
+          cat <<'__ZIZMOR_VALIDATE_PY__' > "${SCRIPT}"
+          #!/usr/bin/env python3
+          """Fail if a repo-local zizmor.yml violates Grafana shared-workflows policy."""
+
+          from __future__ import annotations
```
zerok: Where is that needed?

.github/workflows/reusable-zizmor.yml
```diff
@@ -388,6 +390,188 @@ jobs:
...
+              if not isinstance(unpinned, dict):
+                  _fail(path, "`rules.unpinned-uses` must be a mapping")
+                  return
+
+              if "disable" in unpinned:
+                  _fail(path, "forbidden key `rules.unpinned-uses.disable`")
+
+              cfg = unpinned.get("config")
+              if cfg is None:
+                  return
```
zerok: While the current implementation of the early returns appears to be correct, this is very fragile. Please at least create unittests here or move the whole step into a standalone action where you can have proper testing.

.github/workflows/reusable-zizmor.yml
mikefat: Removed!

.github/workflows/reusable-zizmor.yml
mikefat: Cursor was concerned depending on the python version, confirmed it's not needed

ezebunandu (commented):
It would be nice to add a testcase for catching multiple violations in one config, I think. Also nit, but could add tests for the other two forbidden audits.

Test cases to document that the silent pass for when there's no config file or no rules in the config file would be nice too, but that's a nittier nit -:)

.github/workflows/reusable-zizmor.md
```diff
@@ -157,6 +157,23 @@ uses: actions/checkout@v3 # zizmor: ignore[artipacked]
...
+
+- `always-use-default-config` is `true` (only the Grafana default from this repository is used), or
+- There is no repo-local config file, or
+- The run uses the fetched default via `--config` (no repo-local file in play).
+
+The policy rejects configs that:
+
+- Define any of these audit blocks under `rules` (no `disable`, `ignore`, or `config` is allowed — remove the block entirely): `insecure-commands`, `template-injection`, `impostor-commit`, `known-vulnerable-actions`, and `ref-confusion`.
+- Set `rules.unpinned-uses.disable`.
+
```
kleimkuhler, line 175: Missing a rejection case mentioned in the PR description 

```
- Set `rules.unpinned-uses.config.policies` with a `"*": any` universal entry.
```

kleimkuhler (commented):
Overall looking good. Just a doc comment right now

ezebunandu (approved):
LGTM.

actions/validate-zizmor-config/action.yml
```diff
@@ -0,0 +1,26 @@
...
+description: >-
+  Enforces Grafana policy on a repo-local zizmor.yml (used from reusable-zizmor before running zizmor).
+inputs:
+  config_path:
+    description: Path to the zizmor config file relative to the workspace (e.g. zizmor.yml).
+    required: true
+  pyyaml_version:
+    description: PyYAML version passed to uv run --with.
+    required: false
+    # renovate: datasource=pypi depName=pyyaml
```
zerok: I'm not 100% certain that this will work. The renovate config here currently uses the following pattern:

```
    {
      customType: "regex",
      managerFilePatterns: [
        "/(?:^|/)\\.github/(?:workflows|actions)/.+\\.ya?ml$/",
        "/(?:^|/)action\\.ya?ml$/",
      ],
      matchStrings: [
        "# renovate: datasource=(?<datasource>[a-z-.]+?) depName=(?<depName>[^\\s]+?)(?: (?:lookupName|packageName)=(?<packageName>[^\\s]+?))?(?: versioning=(?<versioning>[^\\s]+?))?(?: extractVersion=(?<extractVersion>[^\\s]+?))?\\s+[A-Za-z0-9_-]+[_-](?:VERSION|version)\\s*[:=]\\s*[\"']?(?<currentValue>[^\"'@\\n]+)(?:@(?<currentDigest>sha256:[a-f0-9]+))?[\"']?",
      ],
    },

```

Based on that the line including the version number would need to have a format like this:

```
SOMETHIGN_VERSION: "1.2.3"
```

Instead, you could set the input to be empty by default and then set the "default" version down below when you inject the PYYAML_VERSION into the process.

.github/workflows/test-validate-zizmor-config.yml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Setup UV
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
+        with:
+          enable-cache: true
+
+      - name: Unit tests
+        run: |
+          cd actions/validate-zizmor-config
+          uv run --with pyyaml==6.0.2 python3 -m unittest discover -v
```
zerok: Please also auto-update this version

zerok (changes requested):

actions/validate-zizmor-config/action.yml
mikefat: Thanks for all your comments! Just pushed some more changes

zerok (approved):

PR #1893, opened 2026-04-22
feat(zizmor): optional vendored-path excludes (security-appsec#326)

isaiah-grafana:
This adds optional **`.github/zizmor-collection-ignore`**: if it’s missing, zizmor still runs on `\.` like before. If it’s there, we `find` workflow/action/dependabot paths with your prefixes pruned out (so vendored trees with nested workflows don’t get pulled in). Big lists get split into batches of 400 and SARIF is merged with `jq`.

Tracks https://github.com/grafana/security-appsec/issues/326

**After this merges:** pin `grafana/security-github-actions` `self-zizmor.yaml` to `grafana/shared-workflows/.github/workflows/reusable-zizmor.yml@<this merge SHA>`, then `deployment_tools` PR adds the ignore file.

Details are in `.github/workflows/reusable-zizmor.md`.


actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+            extra_args=extra_args,
+            targets=targets,
+        )
+        rc = _run(cmd, stdout=args.out)
+        if rc == 1:
+            return 1
+        return 0
+
+    runner_temp = Path(os.environ["RUNNER_TEMP"])
+    tmp = Path(tempfile.mkdtemp(prefix="zizmor-sarif-", dir=runner_temp))
```
zerok, line 168: You can also use `tempfile.TemporaryDirectory` as context manager to clean up for you.

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
+#!/usr/bin/env python3
```
zerok, line 1: I'm not sure if replacing 10 lines of Bash with 300 lines of Python that actually does more is what we want here 🙂

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+    for raw in paths_file.read_text(encoding="utf-8").splitlines():
+        line = raw.strip()
+        if not line:
+            continue
+        out.append(line)
+    return out
+
+
+def _zizmor_cmd(
+    *,
```
zerok, line 37: This whole helper function IMO adds very little benefit to just doing the call directly with something like subprocess.check_output et al.

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+    min_severity = os.environ["MIN_SEVERITY"]
+    min_confidence = os.environ["MIN_CONFIDENCE"]
+    cache_dir = Path(os.environ["ZIZMOR_CACHE_DIR"])
+
+    cfg_raw = os.environ.get("ZIZMOR_CONFIG_PATH", "").strip()
+    config_path = Path(cfg_raw) if cfg_raw else None
+
+    extra_args = _parse_extra_args(os.environ.get("ZIZMOR_EXTRA_ARGS"))
+
+    use_explicit = os.environ.get("USE_EXPLICIT_PATHS", "").strip().lower() == "true"
```
zerok, line 128: This whole thing is done twice. Perhaps you should do that in `main` and then just hand the configuration down?

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+                        zizmor_exit_code = rc
+
+        gh_fh.write("EOF\n")
+        gh_fh.write(f"zizmor-exit-code={zizmor_exit_code}\n")
+
+    return 0
+
+
+def main(argv: list[str]) -> int:
+    if len(argv) < 2:
```
zerok, line 287: argparse can do subcommands: https://docs.python.org/dev/library/argparse.html#subcommands

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+        if not use_explicit:
+            zizmor_exit_code = run_plain(["."], gh_fh=gh_fh)
+        else:
+            paths_list = Path(os.environ["PATHS_LIST"])
+            targets = _read_nonempty_lines(paths_list)
+            if not targets:
+                zizmor_exit_code = 0
+            elif len(targets) <= args.batch_size:
+                zizmor_exit_code = run_plain(targets, gh_fh=gh_fh)
+            else:
```
zerok, line 267: Shouldn't `range` be enough here instead of manual handling of the edge-cases?

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+import json
+import os
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+from typing import Any, TextIO
+
+
+def _truthy_env(name: str) -> bool:
```
zerok, line 21: You use that in only a single place. Personally, I'd just move the check for the input parameters into main and not need that kind of helper.

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+
+    if len(docs) == 1:
+        merged = docs[0]
+    else:
+        schema = docs[0].get("$schema")
+        version = docs[0].get("version")
+        runs: list[Any] = []
+        for doc in docs:
+            r = doc.get("runs")
+            if isinstance(r, list):
```
zerok, line 98: What would be another case here? `null` or that it's simply not defined at all? For the latter, you could use `do.get('runs', [])` in the line above

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+            for line in p.stdout:
+                gh_fh.write(line)
+        finally:
+            stdout_io = p.stdout
+            close = getattr(stdout_io, "close", None)
+            if callable(close):
+                try:
+                    close()
+                except OSError:
+                    pass
```
zerok, line 251: This whole thing feels like overkill to me

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,303 @@
...
+        rc = _run(cmd, stdout=args.out)
+        if rc == 1:
+            return 1
+        return 0
+
+    runner_temp = Path(os.environ["RUNNER_TEMP"])
+    tmp = Path(tempfile.mkdtemp(prefix="zizmor-sarif-", dir=runner_temp))
+    try:
+        parts: list[Path] = []
+        for idx in range(0, len(targets), args.batch_size):
```
zerok, line 171: You could also use `itertools.batched` for most of that logic :)

actions/zizmor-collection-paths/collection_paths.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python3
+"""Build explicit zizmor scan paths when ignore prefixes are configured."""
+
+import argparse
+import os
+import sys
+from pathlib import Path
+
+
+def _lines_after_sed_filter(content: str) -> list[str]:
```
zerok, line 10: Sed has nothing to do with this anymore 🙂

actions/zizmor-collection-paths/collection_paths.py
```diff
@@ -0,0 +1,144 @@
...
+
+def collect_paths(repo_root: Path, prefixes: list[str], paths_out: Path) -> None:
+    repo_root = repo_root.resolve()
+    paths_out.parent.mkdir(parents=True, exist_ok=True)
+    excluded_roots = [(repo_root / p).resolve() for p in prefixes]
+
+    collected: list[str] = []
+    for current_root, dirs, files in os.walk(repo_root, topdown=True):
+        current_path = Path(current_root).resolve()
+        dirs[:] = [
```
zerok, line 77: Why don't you just create a new variable here instead of re-using `dirs`?

actions/zizmor-collection-paths/collection_paths.py
```diff
@@ -0,0 +1,144 @@
...
+
+    s = s.removeprefix("/")
+
+    # Support a couple common spellings people use in ignore files, but reject
+    # arbitrary glob patterns (we only exclude directory prefixes).
+    while True:
+        before = s
+        s = s.removesuffix("**").removesuffix("/*")
+        while s.endswith("/"):
+            s = s[:-1]
```
zerok, line 36: `s.rstrip('/')`

PR #1908, opened 2026-04-28
feat: output Argo workflow URL in summary

### LINKED ISSUE #1907: trigger-argo-workflow could show the Argo Workflow URL in the summary
The trigger-argo-workflow action results in an URL for the Argo workflow that was submitted.

It would be nice if the action included that information in the summary, so that the user doesn't have to go fishing for it in the workflow's output.

mem:
The result of trigger-argo-workflow is a URL for the corresponding Argo Workflow that was submitted. Add another step to the action so that this information is appended to the workflow's summary if the user requestes it.

Fixes: https://github.com/grafana/shared-workflows/issues/1907

### DIFF
--- a/actions/trigger-argo-workflow/action.yaml
+++ b/actions/trigger-argo-workflow/action.yaml
@@ -29,6 +29,11 @@ inputs:
     description: |
       The log level to use. Choose from `debug`, `info`, `warn` or `error`. Defaults to `info`.
     default: info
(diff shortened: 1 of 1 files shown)


dblinkhorn (approved):

mem:
@dblinkhorn thanks! Could you please merge this? I don't have enough access for that.

PR #1945, opened 2026-05-19
feat(zizmor): optional vendored-path excludes

isaiah-grafana:
## Summary

Repos can opt out of zizmor scanning under vendored trees (e.g. `ksonnet/vendor`) that ship upstream `.github/workflows/` copies we do not maintain.

Adds optional **`.github/zizmor-collection-ignore`** (one directory prefix per line). Wired in `reusable-zizmor.yml` via `actions/zizmor-collection-paths/`.

Tracking: https://github.com/grafana/security-appsec/issues/401

## Behavior

| Ignore file | What zizmor does |
|-------------|------------------|
| Missing, empty, or comments only | Scan `.` — **same as today** |
| Has valid prefixes | Scan only workflow-related paths **outside** ignored trees; merge SARIF from batched runs |

Prefix rules: no `..` or absolute paths; trailing `/**` / `/*` stripped. Details in `reusable-zizmor.md`.

Helper scripts run from the OIDC-pinned `shared-workflows` checkout (`_shared-workflows-zizmor/`); that checkout is never scanned.

## Tests

- **21 unit tests** — `actions/zizmor-collection-paths/` (`python3 -m unittest discover -v`)
- CI — `.github/workflows/test-zizmor-collection-paths.yml`

## E2E (no merge required)

Pilot pins this branch at `a0ded699096e4939468a229648b0268340a90c99` in `security-github-actions` `self-zizmor.yaml` (org ruleset uses the PR workflow when the PR is in that repo).

| | |
|---|---|
| PR | https://github.com/grafana/security-github-actions/pull/155 |
| Ruleset run | https://github.com/grafana/security-github-actions/actions/runs/26129732548 |

`vendor-fixture/` is in the ignore file; bait workflow under it was **not** scanned (`USE_EXPLICIT_PATHS=true` in logs).

## After merge

1. Bump `grafana/security-github-actions` `self-zizmor.yaml` to this merge SHA (org ruleset picks it up).
2. Repos add `.github/zizmor-collection-ignore` with their vendored prefixes when ready.

### DIFF
--- a/.github/workflows/reusable-zizmor.yml
+++ b/.github/workflows/reusable-zizmor.yml
@@ -219,14 +219,10 @@ jobs:
 
             const AUDIENCE = 'zizmor-job-workflow-ref';
 
-            // Fork PRs (and some other contexts) have no OIDC token. Detect that
-            // up front and use fallback without throwing, so the step succeeds
-            // and the log stays clear.
+            // Fork PRs have no OIDC token, so we can't detect the workflow ref. Use main.
             const runtimeUrl = process.env['ACTIONS_ID_TOKEN_REQUEST_URL'];
             if (!runtimeUrl) {
-              core.info(
-                "No OIDC token available (e.g. pull request from a fork). Using fallback: grafana/shared-workflows@main for default config."
-              );
+              core.info("No OIDC token (e.g. fork PR). Using grafana/shared-workflows@main for default config.");
               core.setOutput('owner', 'grafana');
               core.setOutput('repo', 'shared-workflows');
               core.setOutput('sha', 'main');
@@ -298,6 +294,31 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           persist-credentials: false
+
+      # .github/zizmor-collection-ignore (optional) — see reusable-zizmor.md
+      - name: Checkout shared-workflows (zizmor helper scripts)
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          repository: ${{ needs.job-workflow-ref.outputs.owner }}/${{ needs.job-workflow-ref.outputs.repo }}
+          ref: ${{ needs.job-workflow-ref.outputs.sha }}
+          path: _shared-workflows-zizmor
+          persist-credentials: false
+
+      - name: Zizmor paths + helper_root
+        id: zizmor-helper
+        shell: bash
+        run: |
+          set -euo pipefail
+          ROOT="${GITHUB_WORKSPACE}/_shared-workflows-zizmor/actions/zizmor-collection-paths"
+          if [ ! -f "${ROOT}/run_zizmor.py" ]; then
+            echo "::error::zizmor helper missing at ${ROOT} (pin reusable-zizmor.yml to a SHA that includes actions/zizmor-collection-paths)." >&2
+            exit 1
+          fi
+          echo "helper_root=${ROOT}" >> "${GITHUB_OUTPUT}"
+          python3 "${ROOT}/collection_paths.py" \
+            --repo-root "${GITHUB_WORKSPACE}" \
+            --ignore-file "${GITHUB_WORKSPACE}/.github/zizmor-collection-ignore" \
+            --paths-out "${RUNNER_TEMP}/zizmor-explicit-inputs.txt"
 
       - name: Restore config from cache
         id: cache-config
@@ -430,22 +451,12 @@ jobs:
         env:
           ZIZMOR_CONFIG_PATH: ${{ steps.setup-config.outputs.zizmor-config }}
           ZIZMOR_CACHE_DIR: ${{ runner.temp }}/.cache/zizmor
-        shell: sh
+          USE_EXPLICIT_PATHS: ${{ steps.zizmor-helper.outputs.use_explicit_paths }}
+          PATHS_LIST: ${{ steps.zizmor-helper.outputs.paths_list }}
+        shell: bash
         run: |
-          uvx zizmor@"${ZIZMOR_VERSION}" \
-            --format sarif \
-            --min-severity "${MIN_SEVERITY}" \
-            --min-confidence "${MIN_CONFIDENCE}" \
-            --cache-dir "${ZIZMOR_CACHE_DIR}" \
-            ${ZIZMOR_CONFIG_PATH:+--config "${ZIZMOR_CONFIG_PATH}"} \
-            ${RUNNER_DEBUG:+"--verbose"} \
-            ${ZIZMOR_EXTRA_ARGS:+${ZIZMOR_EXTRA_ARGS}} \
-            . \
-            > results.sarif
-          exit_code=$?
-          # Only fail on exit code 1 (zizmor crash). 0 = no findings, 10-14 = findings by severity.
-          if [ "${exit_code}" -eq 1 ]; then exit 1; fi
-          exit 0
+          set -euo pipefail
+          python3 "${{ steps.zizmor-helper.outputs.helper_root }}/run_zizmor.py" sarif --batch-size 400 --out results.sarif
 
       - name: Upload artifact
         if: ${{ !cancelled() && hashFiles('results.sarif') != '' }}
@@ -495,12 +506,13 @@ jobs:
           REPO_LOCAL_ZIZMOR_CONFIG: ${{ steps.setup-config.outputs['repo-local-zizmor-config'] }}
           ZIZMOR_CONFIG_PATH: ${{ steps.setup-config.outputs.zizmor-config }}
           ZIZMOR_CACHE_DIR: ${{ runner.temp }}/.cache/zizmor
+          USE_EXPLICIT_PATHS: ${{ steps.zizmor-helper.outputs.use_explicit_paths }}
+          PATHS_LIST: ${{ steps.zizmor-helper.outputs.paths_list }}
         run: |
-          set -o pipefail
-
-          echo "zizmor-results<<EOF" >> "${GITHUB_OUTPUT}"
+          set -euo pipefail
 
           if [ "${VALIDATE_CONFIG_OUTCOME}" = "failure" ]; then
+            echo "zizmor-results<<EOF" >> "${GITHUB_OUTPUT}"
             echo "Repo-local zizmor config policy validation failed for ${REPO_LOCAL_ZIZMOR_CONFIG}."
             if [ -f results.sarif ]; then
               python3 - <<'PY' | tee -a "${GITHUB_OUTPUT}"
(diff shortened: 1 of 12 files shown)


guicaulada (changes requested):
Superseded by the review below — reposted to fix suggestion anchoring.

.github/workflows/reusable-zizmor.yml
```diff
@@ -299,6 +295,31 @@ jobs:
...
+          ROOT="${GITHUB_WORKSPACE}/_shared-workflows-zizmor/actions/zizmor-collection-paths"
+          if [ ! -f "${ROOT}/run_zizmor.py" ]; then
+            echo "::error::zizmor helper missing at ${ROOT} (pin reusable-zizmor.yml to a SHA that includes actions/zizmor-collection-paths)." >&2
+            exit 1
+          fi
+          echo "helper_root=${ROOT}" >> "${GITHUB_OUTPUT}"
+          python3 "${ROOT}/collection_paths.py" \
+            --repo-root "${GITHUB_WORKSPACE}" \
+            --ignore-file "${GITHUB_WORKSPACE}/.github/zizmor-collection-ignore" \
+            --paths-out "${RUNNER_TEMP}/zizmor-explicit-inputs.txt"
```
guicaulada: #2036/#2032 removed shared-workflows checkouts from callers' workspaces (working-tree pollution, git side effects). This one is also *scanned*: default mode runs zizmor on `.`, which includes this checkout — `_SKIP_TOP_LEVEL_DIRS` only helps in explicit-path mode.

Reusable workflows have no `github.action_path`, so wrap the scripts in a composite action and call it as a pinned sibling ref:

```yaml
      - name: Collect zizmor paths
        id: zizmor-helper
        uses: grafana/shared-workflows/actions/zizmor-collection-paths@<sha> # zizmor-collection-paths/v1.0.0
```

The `action.yml` just runs `collection_paths.py` from `${GITHUB_ACTION_PATH}` and outputs `use_explicit_paths`, `paths_list`, and `helper_root=${GITHUB_ACTION_PATH}`. It runs from the actions cache, so nothing lands in the workspace. Also removes the OIDC ref dependency for the helper, the "helper missing" error case, and `_SKIP_TOP_LEVEL_DIRS`; adds the action to release-please like its siblings. Update `reusable-zizmor.md` to match.

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,162 @@
...
+    else:
+        runs = []
+        for doc in docs:
+            runs.extend(doc.get("runs") or [])
+        merged = {
+            "$schema": docs[0].get("$schema"),
+            "version": docs[0].get("version"),
+            "runs": runs,
+        }
+    dst.write_text(json.dumps(merged), encoding="utf-8")
```
guicaulada, line 98: GitHub rejects SARIF files with multiple runs for the same tool + category ([changelog](https://github.blog/changelog/2025-07-21-code-scanning-will-stop-combining-multiple-sarif-runs-uploaded-in-the-same-sarif-file/)), so >400 inputs = failed upload. Merge into a single run and remap `ruleIndex` (each batch indexes its own `rules`). Add a test asserting `len(merged["runs"]) == 1`.

```suggestion
def _merge_sarif_parts(parts: list[Path], dst: Path) -> None:
    docs = [json.loads(p.read_text(encoding="utf-8")) for p in parts]
    runs = [run for doc in docs for run in (doc.get("runs") or [])]
    # Code scanning rejects multiple runs with the same tool and category:
    # fold every batch into the first run, remapping ruleIndex.
    if len(runs) > 1:
        base = runs[0]
        rules = base.setdefault("tool", {}).setdefault("driver", {}).setdefault("rules", [])
        index_by_id = {r.get("id"): i for i, r in enumerate(rules)}
        for run in runs[1:]:
            run_rules = run.get("tool", {}).get("driver", {}).get("rules") or []
            for result in run.get("results") or []:
                rule_id = result.get("ruleId")
                if rule_id is not None and rule_id not in index_by_id:
                    old = result.get("ruleIndex")
                    rule = run_rules[old] if old is not None and old < len(run_rules) else {"id": rule_id}
                    index_by_id[rule_id] = len(rules)
                    rules.append(rule)
                if "ruleIndex" in result and rule_id in index_by_id:
                    result["ruleIndex"] = index_by_id[rule_id]
                base.setdefault("results", []).append(result)
        runs = [base]
    merged = {
        "$schema": docs[0].get("$schema"),
        "version": docs[0].get("version"),
        "runs": runs,
    }
    dst.write_text(json.dumps(merged), encoding="utf-8")
```

actions/zizmor-collection-paths/run_zizmor.py
```diff
@@ -0,0 +1,162 @@
...
+                    return 1
+                fh.write(proc.stdout.read())
+                rc = proc.wait()
+                if rc == 1:
+                    print("zizmor crashed; see output above.", file=sys.stderr)
+                    return 1
+                code = max(code, rc)
+        fh.write("EOF\n")
+        fh.write(f"zizmor-exit-code={code}\n")
+    return 0
```
guicaulada: The `return 1` paths leave the `<<EOF` heredoc unterminated in `GITHUB_OUTPUT`, so the runner fails with "matching delimiter not found" and masks the real error. Close it in a `finally`; `subprocess.run` also drops the `stdout is None` branch.

```suggestion
def _run_plain(paths: list[str] | None, batch: int) -> int:
    gh = os.environ.get("GITHUB_OUTPUT")
    if not gh:
        print("GITHUB_OUTPUT is not set", file=sys.stderr)
        return 2

    with open(gh, "a", encoding="utf-8") as fh:
        fh.write("zizmor-results<<EOF\n")
        code = 0
        try:
            if paths is not None:
                for chunk in batched(paths, batch):
                    proc = subprocess.run(
                        _zizmor_cmd("plain", list(chunk)),
                        stdout=subprocess.PIPE,
                        stderr=subprocess.STDOUT,
                        text=True,
                        check=False,
                    )
                    fh.write(proc.stdout)
                    if proc.returncode == 1:
                        print("zizmor crashed; see output above.", file=sys.stderr)
                        return 1
                    code = max(code, proc.returncode)
        finally:
            fh.write("EOF\n")
        fh.write(f"zizmor-exit-code={code}\n")
    return 0
```

.github/workflows/reusable-zizmor.yml
```diff
@@ -619,7 +607,6 @@ jobs:
 
   grafana-bench:
     name: Send zizmor metrics to Prometheus via Grafana Bench
```
guicaulada, line 593: `continue-on-error: true` was dropped — unrelated and undocumented. Bench failures will now fail consumers' runs. Restore or explain.

```suggestion
    name: Send zizmor metrics to Prometheus via Grafana Bench
    continue-on-error: true
```

.github/workflows/reusable-zizmor.yml
```diff
@@ -430,22 +451,12 @@ jobs:
         env:
           ZIZMOR_CONFIG_PATH: ${{ steps.setup-config.outputs.zizmor-config }}
           ZIZMOR_CACHE_DIR: ${{ runner.temp }}/.cache/zizmor
-        shell: sh
+          USE_EXPLICIT_PATHS: ${{ steps.zizmor-helper.outputs.use_explicit_paths }}
+          PATHS_LIST: ${{ steps.zizmor-helper.outputs.paths_list }}
+        shell: bash
```
guicaulada, line 441: Pass `helper_root` via `env:` instead of interpolating into `run:` — matches the rest of the file and keeps zizmor's template-injection audit quiet. Same in the plain-output step at line 533.

```suggestion
          USE_EXPLICIT_PATHS: ${{ steps.zizmor-helper.outputs.use_explicit_paths }}
          PATHS_LIST: ${{ steps.zizmor-helper.outputs.paths_list }}
          HELPER_ROOT: ${{ steps.zizmor-helper.outputs.helper_root }}
        shell: bash
```

.github/workflows/reusable-zizmor.yml
```diff
@@ -430,22 +451,12 @@ jobs:
...
-            ${RUNNER_DEBUG:+"--verbose"} \
-            ${ZIZMOR_EXTRA_ARGS:+${ZIZMOR_EXTRA_ARGS}} \
-            . \
-            > results.sarif
-          exit_code=$?
-          # Only fail on exit code 1 (zizmor crash). 0 = no findings, 10-14 = findings by severity.
-          if [ "${exit_code}" -eq 1 ]; then exit 1; fi
-          exit 0
+          set -euo pipefail
+          python3 "${{ steps.zizmor-helper.outputs.helper_root }}/run_zizmor.py" sarif --batch-size 400 --out results.sarif
```
guicaulada: …and use the env var here:

```suggestion
          set -euo pipefail
          python3 "${HELPER_ROOT}/run_zizmor.py" sarif --batch-size 400 --out results.sarif
```

actions/zizmor-collection-paths/collection_paths.py
```diff
@@ -0,0 +1,153 @@
...
+        if p:
+            prefixes.append(p)
+    return prefixes
+
+
+def excluded(path: Path, roots: list[Path]) -> bool:
+    return any(path == r or r in path.parents for r in roots)
+
+
+def want_file(path: Path, repo_root: Path) -> bool:
```
guicaulada, line 56: This mirrors zizmor's input collection, which can change across version bumps — add a docstring noting which zizmor version it matches so bump PRs re-check. Also: symlinked workflows resolve outside `repo_root` and get silently dropped in explicit mode.

actions/zizmor-collection-paths/collection_paths.py
```diff
@@ -0,0 +1,153 @@
...
+def parse_prefixes_from_ignore(content: str) -> list[str]:
+    prefixes = []
+    for line in content.splitlines():
+        line = line.strip()
+        if not line or line.lstrip().startswith("#"):
+            continue
+        p = normalize_prefix_line(line)
+        if p:
+            prefixes.append(p)
+    return prefixes
```
guicaulada, line 49: Nit: blank/comment checks duplicate `normalize_prefix_line`.

```suggestion
def parse_prefixes_from_ignore(content: str) -> list[str]:
    prefixes = []
    for line in content.splitlines():
        p = normalize_prefix_line(line)
        if p:
            prefixes.append(p)
    return prefixes
```

.github/workflows/reusable-zizmor.yml
```diff
@@ -558,12 +544,14 @@ jobs:
         if: ${{ !cancelled() && steps.upload-sarif.outcome == 'failure' && github.event.pull_request.head.repo.full_name == github.repository }}
         id: hide-comments
         uses: int128/hide-comment-action@8cd375395d4b1630f8b70d17b802501a15d32561 # v1.66.0
+
```
guicaulada: Nit: stray blank lines after `uses:` here, at 554, and in the bench `env:` block; plus the reworded OIDC comment — unrelated noise, please drop.

guicaulada (changes requested):
Solid feature and good test coverage, but two blockers:

1. **Reintroduces the shared-workflows checkout we removed in #2036/#2032** — and it gets scanned: without an ignore file zizmor scans `.`, which now includes `_shared-workflows-zizmor/`. Every consumer without an ignore file starts seeing shared-workflows' workflows/actions in their results, breaking the "missing ignore file = same as today" promise. Make `zizmor-collection-paths` a composite action called via pinned sibling ref instead (inline).
2. **Batched SARIF merge gets rejected by code scanning.** GitHub rejects files with multiple runs for the same tool + category ([changelog](https://github.blog/changelog/2025-07-21-code-scanning-will-stop-combining-multiple-sarif-runs-uploaded-in-the-same-sarif-file/)), so any repo with >400 inputs fails to upload. Merge into a single run (inline).

Smaller items inline: `grafana-bench` lost `continue-on-error`, `_run_plain` can leave an unterminated heredoc, plus a few nits.

.github/workflows/reusable-zizmor.yml
isaiah-grafana: This has been fixed!

actions/zizmor-collection-paths/run_zizmor.py
isaiah-grafana: I folded batches into a single batch and remapped `ruleIndex`

actions/zizmor-collection-paths/run_zizmor.py
isaiah-grafana: This has been fixed in the latest commit

.github/workflows/reusable-zizmor.yml
isaiah-grafana: Yeah, this is fixed. It got accidentally removed while I was resolving a merge conflict.

.github/workflows/reusable-zizmor.yml
isaiah-grafana: Fixed! `helper_root` is now passed via env in both sarif and plain-output steps.

.github/workflows/reusable-zizmor.yml
isaiah-grafana: Updated!

actions/zizmor-collection-paths/collection_paths.py
isaiah-grafana: I added a docstring on `want_file` noting it mirrors zizmor 1.28.0

actions/zizmor-collection-paths/collection_paths.py
isaiah-grafana: This has been fixed!

.github/workflows/reusable-zizmor.yml
isaiah-grafana: I cleaned this up and removed those blanks.

guicaulada (approved):
Thank you for addressing my concerns!

Only one thing left, the pin is a PR-branch commit and its comment says `# zizmor-collection-paths/v0.1.0`, a tag that won't exist until release-please runs after merge.

It works (SHA-pinned actions resolve from any commit in the repo), but after the release we should re-pin to the actual tag SHA so Renovate tracks it and the comment stops lying.

PR #1946, opened 2026-05-21
feat(create-github-app-token): revoke token in post-job step

eloymg:
Add a post-job step that revokes the Vault token, which invalidates the token as soon as the job finishes. If revocation fails, the token still expires naturally via its Vault lease TTL. This invalidates the Github token in cascade.

Validation here: https://github.com/grafana/deployment_tools/pull/592570


psof (approved):
Nice! Work with SecOps to alert on tokens that have not been revoked indicating a potential tamper with this

PR #1947, opened 2026-05-21
fix(zizmor): send results when config file validation fails

mikefat:
When zizmor config validation fails, no sarif report was being sent, and the logs did not end up in loki. This PR fixes that so that we can see the logs when it fails


mikefat:
[Test action](https://github.com/grafana/test-new-zizmor-workflow/pull/1) shows correct output and loki logs

kleimkuhler (approved):

guicaulada (dismissed):

guicaulada (changes requested):
The change looks good overall, but `__pycache__` shouldn't be committed and needs to be removed and added to `.gitignore`.

mikefat:
Whoops! Thanks for the catch. An oversight

.github/workflows/reusable-zizmor.yml
```diff
@@ -392,10 +392,13 @@ jobs:
           cache-dependency-glob: ""
 
       - name: Validate repo-local zizmor config policy
+        id: validate-config
+        continue-on-error: true
         if: ${{ inputs.always-use-default-config == false && steps.setup-config.outputs.zizmor-config == '' && steps.setup-config.outputs['repo-local-zizmor-config'] != '' }}
-        uses: grafana/shared-workflows/actions/validate-zizmor-config@c2b99bc7ec69e11c1c4e4cba5cbc9197f794b683
+        uses: grafana/shared-workflows/actions/validate-zizmor-config@a476b2f125c0d35381b8b20f09b1c5d9ab95cd1c
```
guicaulada: I noticed this references a commit on the PR itself? Should this be part of this PR? Shouldn't this file only be updated on a follow up PR after the change to the validate-zizmor-config actions is released by release-please?

mikefat: Removed commit reference in favor of updating the action later in release-please

guicaulada (approved):

PR #1953, opened 2026-05-26
fix(create-github-app-token): add masking for GitHub token in create_token.sh

korniltsev-grafanista:


korniltsev-grafanista:
@codex review

korniltsev-grafanista:
fixed in https://github.com/grafana/shared-workflows/pull/1948

PR #1954, opened 2026-05-26
feat(zizmor): auto-delete branches with dangerous-triggers findings

svennergr:
## Summary

- Add `auto-delete-dangerous-branches` and `auto-delete-slack-channel-id` inputs to the reusable zizmor workflow (both optional, auto-delete disabled by default).
- On push events to non-default branches, if zizmor SARIF contains `dangerous-triggers` findings, a new job sends a Slack notification and deletes the branch via the API — closing the attack window without waiting for human review.
- Callers must grant `contents: write` and `id-token: write`, and provide a Slack channel ID.
- Documents the new inputs, required permissions, and usage in `reusable-zizmor.md`.

## Test plan

- [x] Verify the SARIF parsing correctly detects `dangerous-triggers` findings (`found=true`) and ignores unrelated findings (`found=false`).
- [x] Verify the Slack payload is valid JSON with correct fields, formatting, and special-character escaping.
- [x] Verify the job fails early with a clear error when `auto-delete-slack-channel-id` is empty.
- [ ] Test end-to-end on a non-default branch with a known dangerous-trigger workflow file.
- [ ] Confirm Slack notification is sent and branch is deleted with `contents: write` and `id-token: write`.
- [ ] Confirm the job is skipped when `auto-delete-dangerous-branches` is `false` (default).
- [ ] Confirm the job is skipped on the default branch even when enabled.


.github/workflows/reusable-zizmor.yml
```diff
@@ -436,6 +455,27 @@ jobs:
           path: results.sarif
           retention-days: 5
 
+      - name: Check for dangerous-triggers findings
+        id: check-dangerous-triggers
+        if: ${{ !cancelled() && hashFiles('results.sarif') != '' }}
+        shell: bash
```
Proximyst, line 461: nit: we could just use `shell: python` here

.github/workflows/reusable-zizmor.yml
```diff
@@ -644,3 +684,105 @@ jobs:
...
+          print("payload<<EOF")
+          print(json.dumps(payload))
+          print("EOF")
+          PY
+
+      - name: Notify Slack
+        uses: grafana/shared-workflows/actions/send-slack-message@eb1fbd807f87aea8f40ff08dc9cd02872cad55b3 # send-slack-message/v2.0.5
+        with:
+          method: chat.postMessage
+          payload: ${{ steps.slack-payload.outputs.payload }}
```
Proximyst, line 800: How about we delete first, send Slack message after? That way we can also post a "hey, we couldn't delete it, here's a link" message to act as an action-required alert?

.github/workflows/reusable-zizmor.yml
svennergr: Hm, I think that makes this part a bit more difficult/annoying:

```sh
          if [ "${DANGEROUS}" -gt 0 ]; then
            echo "found=true" >> "$GITHUB_OUTPUT"
          else
            echo "found=false" >> "$GITHUB_OUTPUT"
          fi
```

.github/workflows/reusable-zizmor.yml
svennergr: Good idea! Changed to that in c053e12

Proximyst (approved):

zerok (approved):

.github/workflows/reusable-zizmor.yml
```diff
@@ -644,3 +684,123 @@ jobs:
...
+        run: |
+          python3 <<'PY' >> "$GITHUB_OUTPUT"
+          import json
+          import os
+
+          channel = os.environ["SLACK_CHANNEL_ID"]
+          repo = os.environ["REPO"]
+          ref_name = os.environ["REF_NAME"]
+          actor = os.environ["ACTOR"]
+          sha = os.environ["SHA"]
```
Proximyst, line 740: To help with re-creating the branch, we could also print the `sha`, and optionally even print some instructions on how to pull it locally (`git fetch <upstream> <40-char SHA>`) if the local copy is deleted?

I don't know how often this is gonna be a problem, but it seems to be somewhat of a concern from the chat on Slack

svennergr: I really hope this workflow will only trigger in the minority of cases, so for now I'll keep this out of this PR for now.

PR #1955, opened 2026-05-26
feat: replace dawidd6/action-download-artifact with first-party GitHub actions

KucicM:
## What

Replaces `dawidd6/action-download-artifact` (the last remaining third-party
artifact-download dependency in this repo) with first-party GitHub actions:

- **New action:** [`download-branch-workflow-artifact`](actions/download-branch-workflow-artifact/README.md)
  — downloads an artifact from the last successful run of a workflow on a given
  branch, using `actions/github-script` (to find the run) + `actions/download-artifact`
  (to download it). Outputs `found`, `run-id`, and `download-path`.
- **Updated:** `component-change-detection` — replaced the `dawidd6` step with
  the same inline `github-script` + `download-artifact` pattern. After this change,
  `dawidd6/action-download-artifact` is no longer referenced anywhere in the repo.

## Why

Removes a third-party dependency for cross-run artifact downloads, reducing the
supply-chain surface. The replacement uses only SHA-pinned first-party GitHub
actions and requires no binary downloads.

## Testing

- **`test-download-branch-workflow-artifact.yaml`** (new) — exercises the action:
  - nonexistent workflow → asserts `found=false` and an empty `run-id`;
  - existing workflow (`pre-commit.yaml`) with a missing artifact name → asserts
    graceful `found=false`.

  Both jobs are read-only (`contents: read`, `actions: read`) and hermetic.
- **`test-component-change-detection.yml`** (existing) — already exercises the
  inline replacement via `previous-tags-source: test-workflow-nonexistent.yml`,
  validating the "no successful run found" fallback.
- The successful-download path requires a prior workflow run with real artifacts;
  it is validated in practice through `component-change-detection`.

## Notes

- The new action ships a `README.md` (inputs/outputs, required permissions, usage
  example) and is registered with release-please at `0.1.0`.
- `component-change-detection` now requires `actions: read` permission to query
  workflow runs and download cross-run artifacts (its test workflow already grants it).


actions/component-change-detection/action.yml
```diff
@@ -42,14 +42,42 @@ runs:
...
+            }
+          } catch (error) {
+            core.setOutput('run-id', '');
+            console.log(`Failed to query workflow runs: ${error.message}`);
+          }
+
     - name: Download previous component tags
-      uses: dawidd6/action-download-artifact@b6e2e70617bc3265edd6dab6c906732b2f1ae151 # v21
+      if: steps.find-previous-run.outputs.run-id != ''
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
```
zerok, line 75: Hey :) Why is the new action not used here?

KucicM: I initially tried using the new action here, but GitHub Actions had trouble resolving references between composite actions in the same repository. It worked within this repository, but failed when used from stack-state-service. Inlining the logic was the simplest and most reliable solution.

zerok: Yes, that's a problem we also faced with a few other actions. Due to that we've now refactored the whole thing to use versioned actions instead of loading them from the file system :) Not critical right now but might be a nice follow-up.

zerok (dismissed):

zerok (approved):

PR #1974, opened 2026-05-28
feat: add dockerfile conftest scanning to docker-build-push-image action

aarogoss:
## Summary
Adds a `conftest` Dockerfile baseline hardening step to the `docker-build-push-image` action.

Prior to building the Docker image, this process first performs a `conftest` policy scan against the `Dockerfile`, checking for hardening requirements.  (These requirements are listed below.)

With this change, a failure in the `conftest` scan will emit warnings in the workflow, indicating the violations and hardening concerns.  If a calling project wishes to _enforce_ these policies (i.e., fail the action if a violation occurs), they can set `enforce-hardened-dockerfile` to `true`.  Doing so will cause the action to fail if any `conftest` policy tests fail.  This property is set to `false` so the default behavior is warning-only.

This commit also introduces a CI workflow which tests the `conftest` Rego policies which are imposed.

## Dockerfile Requirements
- Final image/base
   - Must be distroless or utilize the `scratch` image

- Image references
  - **All** image references ust be pinned to an OCI supported digest hash algorithm (`sha256`, `sha512`, `blake3`)

- Supply-chain/build-time exec
  - `ADD` cannot fetch remote URLs
  - `RUN` cannot pipe remote content to a shell (e.g., `curl ... | bash`, `wget ... | sh`, etc.)

- Runtime profile
  - Final stage must declare a non-root (UID 0) `USER`
  - No use of `sudo`
  - No `EXPOSE` of privileged ports
  - No world-writable modes (`chmod 777` or equivalent)
  - JSON-based `ENRTYPOINT/CMD` form  

- Advisory check (emits warning, not a failure)
  - Warn user if `COPY .` is used in a build stage (risk of shipping source tree into final image) 

Each policy in this commit has a paired `<name>_test.rego` test case, covering both positive and negative cases.  Additionally, a happy-path `Dockerfile` is included as a test case to ensure a known "good" `Dockerfile` passes the policy set.

## Usage
```
  - uses: grafana/shared-workflows/actions/docker-build-push-image@docker-build-push-image/...
    with:
      registries: ...
      push: true
      tags: |
        ${{ github.sha }}
        main
      # fail if `conftest` policies are violated
      enforce-hardened-dockerfile: true
```

## Testing
- [x] All `*_test.rego` test cases passing
- [x] Happy-path Dockerfile test passing
- [x] Manual testing against known "bad" Dockerfile cases emit expected failures


julienduchesne:
Something like https://github.com/hadolint/hadolint could get us 80% of the way there without maintaining rego. The things that aren't in there are very opinionated like "only distroless", not sure I see the value in this one if the end result is zero CVE with something else

aarogoss:
> Something like https://github.com/hadolint/hadolint could get us 80% of the way there without maintaining rego. The things that aren't in there are very opinionated like "only distroless", not sure I see the value in this one if the end result is zero CVE with something else

Hmm, I would challenge the purpose of a distroless image being "zero CVE".  Distroless images help reduce the attack surface of a container (removing shells, OS utils, etc.) and require developers to only include what's strictly needed for their workloads.

I do agree there are very opinionated policies here but I think that's kinda the point with a hardening baseline: we're introducing a minimal standard for developers when they build their Docker images.

That said, I'm definitely open to making changes -- some of these checks are perhaps too restrictive.

I can look at `hadolint`, but I'm hesitant to introduce another third-party dependency, as opposed to leveraging a utility which is already available to our workflows.  Maintaining policies is a bit of work, but it also gives us the greatest freedom to adjust when/where needed.

actions/docker-build-push-image/conftest/policy/distroless.rego
```diff
@@ -0,0 +1,29 @@
+package main
+
+import rego.v1
+
+distroless_prefixes := {
+	"gcr.io/distroless/",
+	"cgr.dev/chainguard/",
+	"cgr.dev/chainguard-private/",
```
terryxsun, line 7: Should we allow chainguard from the Docker Hub repo too or is it preferred to only use the `cgr.dev` one?

aarogoss: The repo refs I've seen internally for chainguard are all from `cgr.dev`, so I'd be inclined to just stick with that for the time being.  We can update if we see support for those from Docker Hub.

zerok:
I'm slightly torn between this being added to a "build" action vs. being it's own "validate" action. Especially when people use the multi-arch reusable workflow, this would be a nice addition, but for the actual build action it self slightly out of place.

dblinkhorn:
@aarogoss is this PR still needed? If so i believe it needs a rebase.

aarogoss:
> @aarogoss is this PR still needed? If so i believe it needs a rebase.

Hey @dblinkhorn.  I don't think anyone is really champing at the bit for this particular change :)  I'm going to close it and it could be revisited as a topic if it seems like it's needed in the future.

PR #2026, opened 2026-06-04
feat: add annotate-coverage action

oleg-kozlyuk-grafana:
## Summary

- Adds a new composite action `annotate-coverage` that highlights uncovered lines in a PR diff using Go coverage data. It parses Go coverage files from a directory, intersects them with a git diff, and emits text, Markdown, or GitHub Actions workflow commands (`::notice file=...,line=...::...`) that GitHub renders as PR annotations on the changed lines — no GitHub API client involved.
- The utility is lifted verbatim from https://github.com/oleg-kozlyuk-grafana/go-canopy
- Ships its own Go binary at `cmd/annotate-coverage`, with the existing canopy internal packages (`coverage`, `diff`, `format`, `github`, `local`) and their full test suites. Supports four diff modes: PR diff (`base-ref` + `commit-sha`), branch-vs-HEAD (`base-ref`), single-commit (`commit-sha`), and working-tree (no inputs).
- Wires the new action into `release-please-config.json` and `.release-please-manifest.json` with an initial version of `0.1.0`, following the same pattern as `actions/go-flaky-tests`.

## Test plan

- [ ] `go test -race -short ./...` passes inside `actions/annotate-coverage/` (verified locally).
- [ ] `go build ./cmd/annotate-coverage` produces a working binary (verified locally; `annotate-coverage version` runs).
- [ ] `prettier --check` passes on `action.yaml`, `README.md`, and the release-please configs (verified locally).
- [ ] Try the action end-to-end on a downstream PR to confirm `::notice` annotations appear in the GitHub UI on the expected lines.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


release-please-config.json
```diff
@@ -196,6 +196,11 @@
       "package-name": "issues-update-project-status",
       "extra-files": ["README.md"],
       "initial-version": "0.1.0"
+    },
```
electron0zero, line 199: can we also add the custom renovate config: https://github.com/grafana/shared-workflows#custom-renovate-config

electron0zero (approved):
lgtm.

release-please-config.json
oleg-kozlyuk-grafana: This is for receiving repos that use these actions?

electron0zero (approved):

ricky-undeadcoders (commented):
Nice work — clean structure and great test coverage. I ran it locally: `go test -race -short ./...` passes in all five packages, and the binary builds and runs fine. It follows the same pattern as `go-flaky-tests`, and building into `RUNNER_TEMP` keeps the user's checkout clean.

A couple of notes.

**Should fix:**

- `internal/coverage/analysis.go:38` — matching a coverage file to a diff file by suffix is too loose. It can match the wrong file: a coverage profile for `.../myhandler.go` would match a diff entry for `handler.go`, because `myhandler.go` ends with `handler.go`. So coverage can get attributed to the wrong file.

  Line 41 already has the correct check (it requires a `/` right before the name, so `handler.go` only matches a real path segment). But that line never runs — line 38 matches everything line 41 would, and returns first. So the safe check is effectively dead code.

  Fix: drop the loose check on line 38 and keep the boundary one, plus an exact match for files at the repo root:

  ```go
  if profileFile == diffFile || strings.HasSuffix(profileFile, "/"+diffFile) {
      return diffFile, addedLines, true
  }
  ```

  One more thing: this loops over a map, and Go map order is random. If two diff files both match, which one wins is nondeterministic. The fix above makes matches stricter, which helps, but worth keeping in mind.

**Optional: E2E CI Test**

`test-annotate-coverage.yaml` only runs `go test`. Some of the other `test-*` workflows in this repo also run the action itself and check its output — see [`test-component-change-detection.yml`](https://github.com/grafana/shared-workflows/blob/main/.github/workflows/test-component-change-detection.yml), which does `uses: ./actions/component-change-detection` and then validates the outputs. Right now nothing checks the action wiring end to end (inputs → command-line flags → the `::notice` output GitHub renders). Adding this is good practice to future proof against regressions, but also to demonstrate that it currently works end-to-end.

It's a little more setup here, since the action computes the diff from live git state rather than taking a diff as input — so a meaningful test needs a small git repo plus a coverage file to intersect. If that's too much for this PR, a screenshot of annotations on a real downstream PR would at least confirm GitHub renders the output.

oleg-kozlyuk-grafana:
Addressed the review comments, please take another look

ricky-undeadcoders (approved):

PR #2028, opened 2026-06-08
feat(create-github-app-token)!: rewrite as node24 action with post-job revocation

eloymg:
Rewrite create-github-app-token as node24 action and add post-job token revocation. Please review it carefully because is a complete rewrite.


actions/create-github-app-token/lib.js
```diff
@@ -0,0 +1,138 @@
...
+const crypto = require("node:crypto");
+
+// `::add-mask::` registers `value` as a secret so subsequent log lines that
+// echo it are redacted. Called defensively before logging anything derived
+// from a secret.
+const setSecret = (value) => {
+  if (value) {
+    console.log(`::add-mask::${value}`);
+  }
+};
```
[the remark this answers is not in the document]
eloymg: false positive

actions/create-github-app-token/lib.js
```diff
@@ -0,0 +1,138 @@
...
+    throw new Error("GITHUB_STATE is not set; cannot save state.");
+  }
+  writeKvFile(file, name, value, "ghastate");
+};
+
+const getState = (name) => process.env[`STATE_${name}`] || "";
+
+const info = (message) => console.log(message);
+const warning = (message) => console.log(`::warning::${message}`);
+const error = (message) => console.log(`::error::${message}`);
```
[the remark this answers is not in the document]
eloymg: false positive

Proximyst:
@codex review

psof (commented):
The direction is right: shrinking the GitHub App token's useful lifetime from TTL to ~job duration is a real blast-radius win.

One thing blocks me is that the whole security claim rests on revoking the Vault token actually invalidating the GitHub App token. Lease cascade-revoke is standard Vault behavior, but whether the `github-app-*` secrets engine's revoke propagates to GitHub depends on the plugin.

Please verify with a check (mint -> post -> assert 401 from GitHub). If it doesn't propagate, the README overstates the control.

A non-blocking follow-up: the revoke path is the headline feature but is untested and only emits a ::warning:: on failure. Let's add a detection signal and a couple of status-handling tests so a silent regression doesn't surface during an incident ;)

actions/create-github-app-token/README.md
```diff
@@ -2,6 +2,13 @@
 
 From a `grafana/` org repository, get a ephemeral GitHub API token from a GitHub App using Vault.
 
+The action automatically registers a post-job step that revokes the Vault
+token used to issue the GitHub App token. Revoking the Vault token
+cascade-revokes every lease it created, which invalidates the GitHub App token
+as soon as the job finishes (regardless of whether earlier steps succeeded or
+failed). If revocation fails for any reason, the token still expires naturally
+when its Vault lease TTL elapses.
```
guicaulada, line 10: The post-job revocation claim here ("invalidates the GitHub App token as soon as the job finishes") rests on Vault's `github-app-*` secrets engine propagating lease revocation to GitHub's DELETE /installation/token. Lease cascade-revoke is standard, but plugin-side propagation isn't guaranteed by the cascade alone. Could we verify once (mint → post → assert 401 from GitHub) before merging, given how firm the wording is? If it doesn't propagate, this overstates the control.

actions/create-github-app-token/post.js
```diff
@@ -0,0 +1,126 @@
...
+  setSecret(proxyJwt);
+
+  for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
+    const { status, body } = await revokeOnce({
+      vaultUrl,
+      vaultToken,
+      proxyJwt,
+    });
+
+    if (status >= 200 && status < 300) {
```
guicaulada: The headline feature's status branching (2xx success here, 403/404 early-return at line 101, retry-on-other, network error → status: 0) is untested. Consider extracting the status→action decision into a small pure helper so it can be unit-tested without network mocking.

actions/create-github-app-token/post.js
```diff
@@ -0,0 +1,126 @@
...
+      `Vault token revoke attempt ${attempt}/${MAX_ATTEMPTS} ` +
+        `failed (HTTP ${status}): ${body}`,
+    );
+
+    if (attempt < MAX_ATTEMPTS) {
+      await sleep(RETRY_BASE_DELAY_MS * attempt);
+    }
+  }
+
+  warning(
```
guicaulada: Should a total revoke failure emit a stronger signal than ::warning::? A metric/annotation or distinct marker would make a systemic revoke regression detectable instead of surfacing during an incident. Non-blocking.

actions/create-github-app-token/post.js
```diff
@@ -0,0 +1,126 @@
...
+  } catch (err) {
+    warning(
+      `Failed to mint proxy JWT for Vault revoke-self: ${err.message}. ` +
+        "The Vault token will expire naturally when its TTL elapses.",
+    );
+    return;
+  }
+  setSecret(proxyJwt);
+
+  for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
```
guicaulada: `main.js` uses `lib.retry` but this hand-rolls its loop. It's justified (you need status-specific early-exit for 403/404), but a one-line comment saying why it doesn't reuse `lib.retry` would save the next reader a double-take.

guicaulada (dismissed):
Over all looks good! Got a few suggestions but nothing blocking.

Main concern is validating the mint → revoke → assert 401 flow, if we could do that before getting this merged it would be great.

actions/create-github-app-token/README.md
eloymg: Introduced github token validity check. Only with 3 retries to avoid time locking but I observed that sometimes needs to close 1 min to be revoked

actions/create-github-app-token/post.js
eloymg: We can use the revocation token prometheus metrics for this

eloymg:
@guicaulada @psof Added some of the suggestions. Thanks for the review! 😄

psof (dismissed):
LGTM, thanks for the changes

actions/create-github-app-token/post.js
```diff
@@ -0,0 +1,219 @@
...
+    });
+    const body = await res.text();
+    return { status: res.status, body };
+  } catch (err) {
+    return { status: 0, body: err.message };
+  }
+};
+
+const revokeVaultToken = async ({ vaultUrl, vaultToken, proxyJwt }) => {
+  // using custom retry insted of the lib.js to include
```
NickAnge, line 73: Can you fix the comment here ?

actions/create-github-app-token/main.js
```diff
@@ -0,0 +1,183 @@
...
+const main = async () => {
+  const { permissionSet, vaultInstance, apps, repositoryName } = parseInputs();
+
+  const vaultUrl = `https://vault-github-actions.grafana-${vaultInstance}.net`;
+  const proxyAudience = `vault-github-actions-grafana-${vaultInstance}`;
+  // The Vault auth audience is the same as the Vault URL (Vault's
+  // github-actions-oidc method validates the audience against its configured
+  // value).
+  const vaultAudience = vaultUrl;
+  const workflowRef = process.env.GITHUB_WORKFLOW_REF || "";
```
NickAnge, line 126: If `GITHUB_WORKFLOW_REF` is not specified then workflowRef, is empty which creates a malformed role. What is the expected path here ? Should we fail or is it expected ?

actions/create-github-app-token/main.js
```diff
@@ -0,0 +1,183 @@
...
+  setSecret(proxyJwt);
+
+  const vaultJwt = await retry({ label: "Mint Vault OIDC token" }, () =>
+    fetchIdToken(vaultAudience),
+  );
+  setSecret(vaultJwt);
+
+  // 2) Authenticate with Vault.
+  const vaultToken = await retry({ label: "Vault auth" }, () =>
+    authenticateWithVault({ vaultUrl, proxyJwt, vaultJwt, role }),
```
NickAnge, line 145: Hey 👋 I see that none of the fetch calls set a timeout. If a connection opens but the response never arrives (half-open socket), the promise never settles and retry never gets a chance to kick in, the job just hangs until timeout-minutes kills it.

I wonder if this is worth adding, to avoid this scenario 🤔 Most of the cases it should return an error if it fails

actions/create-github-app-token/main.js
NickAnge: Maybe do a similar to `GITHUB_REPOSITORY` check

NickAnge (commented):
Maybe this is already done, but not specified at the description. Given that this a complete re-write, have we tried run the test workflow on the PR branch before merging. Because the action is used widely.

We use versioning, so it shouldn't affect anyone that do not bump.

Also this is a complete re-write , so i would bump it as major

eloymg:
Sorry for the delay. @NickAnge @psof @guicaulada I added the suggestions and add the commit message as a breaking changes. Thanks!

psof (approved):
Thanks!

NickAnge (approved):
Thank you for addressing my comments

PR #2032, opened 2026-06-09
fix: reference sibling actions directly instead of checkout

guicaulada:
Several composite actions checked out the entire shared-workflows repository just to reference a sibling action via a local path, adding checkout and cleanup steps to every run.

This replaces those checkouts with direct pinned references (`uses: grafana/shared-workflows/actions/<name>@<sha> # <name>/vX.Y.Z`), pinned to the full commit SHA of each action's latest release:

- `send-slack-message` → `get-vault-secrets/v2.0.0`
- `dockerhub-login` → `get-vault-secrets/v2.0.0`
- `argo-lint` → `setup-argo/v1.2.0`
- `issues-update-project-status` → `create-github-app-token/v0.3.1`
- `push-to-gcs` → `login-to-gcs/v0.3.0`

Actions that need the repository contents for source code or scripts (`trigger-argo-workflow`, `docker-build-push-image`, `docker-import-digests-push-manifest`, `lint-pr-title`) keep their checkout.

Note: forks of shared-workflows previously resolved sibling actions from the fork at the calling ref (`github.action_repository`/`github.action_ref`); these actions now always use the pinned upstream release.

### DIFF
--- a/actions/argo-lint/action.yaml
+++ b/actions/argo-lint/action.yaml
@@ -15,22 +15,8 @@ runs:
   using: composite
 
   steps:
-    - name: Checkout
-      env:
-        # In a composite action, these two need to be indirected via the
-        # environment, as per the GitHub actions documentation:
-        # https://docs.github.com/en/actions/learn-github-actions/contexts
-        action_repo: "${{ github.action_repository || 'grafana/shared-workflows' }}"
-        action_ref: ${{ github.action_ref }}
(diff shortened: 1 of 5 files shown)


NickAnge (approved):
I like this change. 

Also if a broken release of a sibling action lands, these consumers aren't impacted until someone explicitly bumps the pin.

PR #2078, opened 2026-06-11
feat(missing-permissions): add PR comments for missing permissions blocks

dblinkhorn:
This PR adds a non-blocking step to `reusable-zizmor.yml` that comments on PRs when any workflow in the calling repo has jobs with no `permissions:` block. Controlled by a new `comment-missing-permissions` input, enabled by default.

## Why

We plan to restrict the org default `GITHUB_TOKEN` permissions to read-only. This gives repos advance notice on every PR so they can declare explicit permissions before workflows relying on the implicit permissions break.

## Changes

- Re-runs zizmor and filters to this one finding type via `awk` (zizmor has no native per-finding-type selection). Never affects scan results, exit codes, or `fail-severity` existing behavior. Lists all affected files, not just those changed by the PR.

**NOTE:** Once this merges it will be necessary for `grafana/security-github-actions` to bump its pinned ref.

## Testing

See zizmor comments below which verify that the changes in this PR preserve the existing high-severity findings/comments behavior and validates that the new missing permissions comment behavior is working as expected. The subsequent commit removing the test workflow also showed both comments being successfully hidden.


guicaulada (approved):
Only thing to raise is that we might want to update the docs on [reusable-zizmor.md](https://github.com/grafana/shared-workflows/blob/main/.github/workflows/reusable-zizmor.md).

Although this step is probably temporary, so maybe editing the readme is not necessary.

dblinkhorn:
> Only thing to raise is that we might want to update the docs on [reusable-zizmor.md](https://github.com/grafana/shared-workflows/blob/main/.github/workflows/reusable-zizmor.md).
> 
> Although this step is probably temporary, so maybe editing the readme is not necessary.

Yeah as part of this PR my opinion is it's not necessary. However, when we move to making this a blocking check in the main run we should definitely update the docs.

PR #2103, opened 2026-06-18
chore(trigger-argo-workflow): bump go version to 1.26

### LINKED ISSUE #2017: trigger-argo-workflow is using go version from go.mod that never gets updated
https://github.com/grafana/shared-workflows/tree/main/actions/trigger-argo-workflow is using the [go version from withing go.mod](https://github.com/grafana/shared-workflows/blob/46f48da11e78ebdba7a8747ae456b11062fac83e/actions/trigger-argo-workflow/action.yaml#L71) - but that seems to not be upgraded and is currently runnign [go1.24.0](https://github.com/grafana/shared-workflows/blob/46f48da11e78ebdba7a8747ae456b11062fac83e/actions/trigger-argo-workflow/go.mod#L3) . 

Should this be just updated in the go.mod or maybe just not put a version so we always get the latest one?

mstoykov:
This is bumped to 1.26 for two reasons:
1. as this isn't a library it doesn't need to support older version so
   it isn't 1.25
2. I am not using patch version so that we always get the latest one
   from setup-go which uses this to decide what to setup

Closes #2017

### DIFF
--- a/actions/trigger-argo-workflow/go.mod
+++ b/actions/trigger-argo-workflow/go.mod
@@ -1,6 +1,6 @@
 module github.com/grafana/shared-workflows/actions/trigger-argo-workflow
 
-go 1.24.0
+go 1.26
 
 require (
 	github.com/cenkalti/backoff/v5 v5.0.3


zerok (approved):

mstoykov:
@zerok I can not merge this - I can not even push to this repo which is why this is from a fork :sweat_smile:

zerok:
> @zerok I can not merge this - I can not even push to this repo which is why this is from a fork 😅

Checking right now if there was a reason behind not giving all GrafanaLabs employees direct write access (it's gated through PRs anyway).

PR #2121, opened 2026-07-01
feat: new signed-commits-info action

zerok:
This action should support PR authors to identify commits that are not yet signed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,7 @@ node_modules/
 
 coverage/
 
+dist/
+!actions/signed-commits-info/dist/
+
 .*.bun-build
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,5 @@ yarn-error.log*
 node_modules/
 
 coverage/
-dist/
 
 .*.bun-build
--- a/actions/signed-commits-info/CLAUDE.md
+++ b/actions/signed-commits-info/CLAUDE.md
@@ -1,0 +1,38 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project
+
+A GitHub Action that runs on `pull_request` events, lists the commits the PR
+introduces over its base ref, and reports those without a fully verified
+signature to `GITHUB_STEP_SUMMARY` and as a comment to the pull request.
+Informational only — never fails the job.
+
+## Toolchain
+
+This project uses **bun**, not npm. Do not introduce `npm`, `yarn`, `pnpm`, or
+`@vercel/ncc` to the workflow.
+
+```sh
+bun install
+bun run build       # bundles src/main.ts → dist/index.js (node target)
+bun run typecheck   # tsc --noEmit (tsconfig is noEmit-only; bun does the bundling)
+```
+
+`dist/index.js` is the action's runtime entry point referenced from
+`action.yml` and **must be committed** so GitHub can execute it without an
+install step.
+
+## Architecture
+
+Single-entrypoint Node action. `src/main.ts`:
+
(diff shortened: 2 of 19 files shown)


release-please-config.json
```diff
@@ -206,6 +206,11 @@
       "package-name": "annotate-coverage",
       "extra-files": ["README.md"],
       "initial-version": "0.1.0"
+    },
+    "actions/annotate-coverage": {
```
NickAnge: It should be `"actions/signed-commits-info"` , no ?

.gitignore
```diff
@@ -13,6 +13,5 @@ yarn-error.log*
 node_modules/
 
 coverage/
-dist/
```
NickAnge, line 16: Removing dist/ globally affects every directory in the repo, not just actions/signed-commits-info/, do we want to scope it only for this one maybe ? Not sure

actions/signed-commits-info/CLAUDE.md
```diff
@@ -0,0 +1,37 @@
+# CLAUDE.md
```
NickAnge, line 1: This is the first `CLAUDE.md` in the repo. Just checking—was this added intentionally, or by mistake?

actions/signed-commits-info/src/main.ts
```diff
@@ -0,0 +1,115 @@
...
+  return octokit.rest.issues.createComment({
+    owner,
+    repo,
+    issue_number: issueNumber,
+    body,
+  });
+}
+
+run().catch((err) =>
+  core.setFailed(err instanceof Error ? err.message : String(err)),
```
NickAnge, line 113: Not very good with ts so might be wrong what i am saying, but I noticed that the catch handler uses core.setFailed() but the README says this action is `informational-only` and never fails the job. If the comment API call fails (e.g., missing permissions, rate limit), this will turn the step red.

actions/signed-commits-info/src/main.ts
```diff
@@ -0,0 +1,115 @@
...
+  } else if (existing) {
+    const body = buildAllVerifiedCommentBody(
+      commits.length,
+      pr.base.ref,
+      pr.head.ref,
+    );
+    await upsertComment(octokit, owner, repo, pr.number, body, existing);
+  }
+
+  return core.summary.write();
```
NickAnge: Minor: core.summary.write() runs after upsertComment(), so if the comment API call fails the step summary is also lost, so the user will not see anything just the failed message at the Actions UI

Do you think its worth to move it before ?

actions/signed-commits-info/src/main.ts
zerok: I'm still not sure what's the best approach here. Personally, I prefer just using https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepscontinue-on-error in the workflow that will then use this step. This way we should at least more easily see system-level issues.

release-please-config.json
zerok: Fixed, thanks :)

.gitignore
zerok: Done :)

actions/signed-commits-info/CLAUDE.md
zerok: Was intended as it makes iterating on the app much easier to have that file in place.

actions/signed-commits-info/src/main.ts
zerok: Good point. This was from a time when I still hoped to combine more of the summary functionality between the two output channels there ^_^ Fixed.

NickAnge (approved):

PR #2144, opened 2026-07-09
feat: add sign-and-attest reusable workflow

dylwylie:
Reusable workflow that cosign-signs (keyless, OIDC) and attaches SLSA build provenance to a digest-pinned container image, then self-verifies both in CI. Adapted from grafana/tempo's `sign-and-attest.yml`, including its GAR auth fix (tempo#7565). First consumer: grafana/tanka.

A reusable workflow rather than a composite action so the Fulcio identity is `grafana/shared-workflows`: one org-wide verification identity, centrally patchable ([GitHub's documented pattern](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-and-reusable-workflows-to-achieve-slsa-v1-build-level-3)). Verifiers bind signatures to a specific repo via the certificate's repository claim; see `sign-and-attest.md`.

Note: the workflow path/name becomes a public verification identity, so renaming it later is a breaking change for all consumers. If you want a different name, now is the time.

Ref: https://github.com/grafana/deployment_tools/issues/620914

### DIFF
--- a/.github/workflows/sign-and-attest.yml
+++ b/.github/workflows/sign-and-attest.yml
@@ -1,0 +1,175 @@
+name: Sign and attest container image
+
+# For a pushed, digest-pinned container image this reusable workflow:
+#   1. Keyless-signs it with cosign via GitHub Actions OIDC (proves the image
+#      was produced by this workflow's identity, on behalf of the calling
+#      repository).
+#   2. Generates and attaches a SLSA build provenance attestation that ties the
+#      image digest to the calling repository's source commit, workflow, and
+#      run (proves *how* and *from what* the image was built).
+#
+# Signature verification identity:
+#   https://github.com/grafana/shared-workflows/.github/workflows/sign-and-attest.yml@<ref>
+#   issuer https://token.actions.githubusercontent.com
+#
+# Because this is a reusable workflow, the identity above is shared by every
+# consuming repository. Verifiers bind a signature to a specific repo with the
+# certificate's repository claim, e.g.:
+#
+#   cosign verify \
+#     --certificate-identity-regexp '^https://github\.com/grafana/shared-workflows/\.github/workflows/sign-and-attest\.yml@' \
+#     --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+#     --certificate-github-workflow-repository grafana/<repo> \
+#     <image>@sha256:...
+#
+#   gh attestation verify oci://<image>@sha256:... \
+#     --repo grafana/<repo> \
+#     --signer-workflow grafana/shared-workflows/.github/workflows/sign-and-attest.yml
+#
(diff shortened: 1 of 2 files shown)


.github/workflows/sign-and-attest.yml
```diff
@@ -0,0 +1,175 @@
...
+            exit 1
+          fi
+          enc="$(printf 'oauth2accesstoken:%s' "$token" | base64 -w0)"
+          cfg="${HOME}/.docker/config.json"
+          mkdir -p "${HOME}/.docker"
+          [ -f "$cfg" ] || echo '{}' > "$cfg"
+          tmp="$(mktemp)"
+          jq --arg reg "$REGISTRY" --arg a "$enc" '.auths[$reg] = {auth: $a}' "$cfg" > "$tmp"
+          mv "$tmp" "$cfg"
+          chmod 0600 "$cfg"
```
zerok, line 126: Should this also be undone at the end of the workflow? (`always()`)

.github/workflows/sign-and-attest.yml
```diff
@@ -0,0 +1,189 @@
...
+
+jobs:
+  sign-and-attest:
+    name: Sign and attest image
+    # Guard on the org, not a single repo: WIF auth to GAR and the verification
+    # contract below only make sense for grafana-owned repos. Callers outside
+    # the org get a clean skip instead of a WIF failure. (Reusable workflows
+    # can't have a workflow-level `if`, so this lives on the job.)
+    if: github.repository_owner == 'grafana'
+    runs-on: ubuntu-24.04
```
Proximyst: Perhaps we should force this onto our self-hosted runners, so that we know they're also watched by Falco and similar? I wouldn't want any of this to be leaked in any compromise event.

.github/workflows/sign-and-attest.yml
```diff
@@ -0,0 +1,189 @@
...
+# (or moving it to another repo) is an org-wide breaking change for anyone
+# verifying. See sign-and-attest.md for the full contract.
+
+on:
+  workflow_call:
+    inputs:
+      image:
+        description: "Digest-pinned image reference, e.g. us-docker.pkg.dev/grafanalabs-global/dockerhub-tanka-prod-mirror/tanka@sha256:..."
+        required: true
+        type: string
```
Proximyst, line 40: Perhaps we should validate that the input image path _is_ a GAR registry? (i.e. `*.pkg.dev/*`)

.github/workflows/sign-and-attest.yml
dylwylie: whoops good shout - thanks!

.github/workflows/sign-and-attest.yml
dylwylie: 👍

Proximyst (approved):

PR #2158, opened 2026-07-10
fix: avoid noisy notice messages

martincostello:
Downgrade `notice` messages` to `debug` to avoid noise in GitHub Actions workflow summaries.

In normal usage, consumers don't need to see things like this in the workflow summaries:

image


zerok (approved):

zerok:
Thank you 🙂

PR #2206, opened 2026-07-24
feat!: export branch-specific SBOM via Socket CLI

ezebunandu:
BREAKING CHANGE: socket-export-sbom now creates a fresh Socket scan for a caller-specified branch instead of reading the repo's existing default-branch scan. Callers must now check out the target branch before invoking this
action (previously no checkout was required).

The custom Go call making API calls was not feasible for this scenario, as we would have needed to handle the messy business of recursively finding manifests files to build the scan/SBOM around. The Socket CLI handles this for us.

I have validated the revised workflow in the [Pysigma-backend-loki repo](https://github.com/grafana/pySigma-backend-loki/blob/main/.github/workflows/sbom-release.yml) and here's [a successful run of the workflow](https://github.com/grafana/pySigma-backend-loki/actions/runs/30375270224/job/90329231571) for reference.


actions/socket-export-sbom/action.yml
```diff
@@ -8,14 +8,21 @@ inputs:
...
-    default: "https://api.socket.dev/v0"
+    default: "https://api.socket.dev/v0/"
   socket_org:
     description: "Socket org name"
     required: true
     default: "grafana"
+  branch:
+    description: "Branch to scan and export the SBOM for. The caller must have already checked out this branch before invoking this action."
+    required: true
   output_file:
```
dblinkhorn, line 19: the readme defines a default "spdx.json" for this input, but it's not specified here in the code

actions/socket-export-sbom/README.md
```diff
@@ -1,17 +1,21 @@
...
-| `output_file`      | `string` | Name of the file to save the socket sbom on the runner.                                                        | `"spdx.json"`                 | false    |
+| Name                     | Type     | Description                                                                                                                                                                                                                          | Default Value                 | Required |
+| ------------------------ | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ----------------------------- | -------- |
+| `socket_api_token`       | `string` | API Key used to authenticate to socket.dev, requires repo: list, repo:read, full-scan:list, report:list scopes                                                                                                                       | `none`                        | true     |
```
dblinkhorn: since the action runs `socket scan create` I think we also need `full-scans:create` listed here as well?

actions/socket-export-sbom/action.yml
```diff
@@ -44,14 +75,44 @@ runs:
...
+          HTTP_CODE=$(curl --silent --output "$DEST" --write-out "%{http_code}" \
+            --header "Authorization: Bearer $SOCKET_API_TOKEN" "$URL") || HTTP_CODE="curl_error"
+
+          if [[ "$HTTP_CODE" == "200" ]] && jq -e . "$DEST" >/dev/null 2>&1; then
+            echo "SBOM ready after ${ATTEMPT} attempt(s)"
+            break
+          fi
+
+          ELAPSED=$(($(date +%s) - START))
+          if ((ELAPSED >= TIMEOUT_SECONDS)); then
```
dblinkhorn, line 113: This timeout is checked only after `curl` returns, so it seems like a stalled request could exceed `TIMEOUT_SECONDS`. Maybe you could leverage curl's `--max-time` flag here?

dblinkhorn (changes requested):

actions/socket-export-sbom/action.yml
ezebunandu: Given how this workflow will be mostly called, we shouldn't require this to be provided, so I've tweaked to just use a sensible default. Thanks

dblinkhorn (approved):

PR #2254, opened 2026-08-19
docs: clarify shared workflow tag formats

zeitlinger:
## What changed

- Use the canonical `<name>/v<semver version>` format consistently in the release documentation.
- Clarify that SHA-pinned examples should retain the complete component tag in their comments.
- Make the example a valid 40-character SHA pin.

## Why

The repository now has both historical hyphenated tags and current slash-prefixed component tags. The documentation should describe the canonical slash format without producing ambiguous or incomplete SHA comments.

## Validation

- `npx --yes prettier@3.9.6 --check README.md`
- `git diff --check`

Related to grafana/shared-workflows#2255

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,11 +60,11 @@ version of the action, even if the action's maintainers release a new version or
 if the tag itself is updated.
 
 Dependabot can update these SHA references when there are new versions. If you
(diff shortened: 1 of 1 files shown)


dblinkhorn (approved):
Thank you!

PR #2258, opened 2026-08-20
fix(socket-export-sbom): run the SPDX export in JS to avoid bot-protection 403s

ezebunandu:
The export-sbom stage now runs in JS (`export-sbom.js`, via `actions/github-script`) instead of `curl`. It was hitting 403s in real use: `curl` sends no `User-Agent`, which we suspect trips the bot protection in front of the Socket API, and the shell loop couldn't tell a 403 from a 404 — it retried auth rejections for the full 180s timeout and then reported them as "not ready". Full retry table is in the [README](actions/socket-export-sbom/README.md#export-behaviour).

Behaviour changes worth a reviewer's attention:

- `401`/`403`/other `4xx` now fail on the first response. `404`/`408`/`429`/`5xx` and network errors still retry with the same 5s-doubling-to-30s backoff.
- Only a body that parses as JSON is written to `output_file`. `curl --output` wrote the 403 challenge page straight to the SBOM destination, so a later step could have uploaded an HTML error page as a release asset.
- `socket_base_url` loses its trailing slash, which had been producing `.../v0//orgs/...`. `exportUrl` strips trailing slashes as well, so a caller passing their own can't reintroduce it.

### DIFF
--- a/actions/socket-export-sbom/export-sbom.js
+++ b/actions/socket-export-sbom/export-sbom.js
@@ -129,7 +129,7 @@ async function fetchSBOM({ fetch, url, token }) {
   } catch (error) {
     return {
       retryable: true,
-      reason: `Socket returned 200 with a body that is not valid JSON: ${error.message}`,
+      reason: `Socket returned 200 with a body that is not valid JSON (${error.message})${bodySnippet(body)}`,
     };
   }
 
--- a/actions/socket-export-sbom/export-sbom.js
+++ b/actions/socket-export-sbom/export-sbom.js
@@ -1,0 +1,220 @@
+// Socket generates the full scan report lazily, so the SPDX export endpoint 404s for a
+// while after `socket scan create` returns. This polls until the report is ready.
+//
+// This replaces a bare `curl` loop that could not tell a 403 from a 404: it retried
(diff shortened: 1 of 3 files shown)


actions/socket-export-sbom/export-sbom.js
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  // Socket has served a 200 holding a partial report while the scan is still being
+  // generated, so the body has to parse before the export counts as done.
+  try {
+    JSON.parse(body);
+  } catch (error) {
+    return {
+      retryable: true,
+      reason: `Socket returned 200 with a body that is not valid JSON: ${error.message}`,
```
ricky-undeadcoders: On line 121 you include the body snippet as part of the error response. Do we care about it here as well?

ricky-undeadcoders (dismissed):
LGTM, I left one comment, but happy for this to go in either way.

actions/socket-export-sbom/export-sbom.js
ezebunandu: Thanks. That sounds useful, so I have made the change.

ricky-undeadcoders (approved):

Issue #1466, opened 2025-10-28
Finalize Docs Generator and Implement CI Check


ricky-undeadcoders:
Work has been started on a script that can parse a composite action or reusable workflow, extract the inputs and outputs defined, and then output those inputs/outputs formatted for either a composite action, reusable workflow, or markdown table.

https://github.com/grafana/shared-workflows/pull/1454

We should look into putting some extra love into this to do the following:

1. Write a check that verifies that `.github/workflows/docker-build-push-multiarch.yml` includes all of the inputs and outputs defined in `actions/docker-build-push-image/action.yml`.
2. Write a check that verifies that all actions have their inputs and outputs documented in the associated README.
3. Write a check to verify that any reusable workflows (in `.github/workflows`) have their inputs/outputs documented in an associated README.

ricky-undeadcoders:
I have 2 PRs open that should complete this task.

**The first (#1454) is a new script** at `scripts/generate-input-output-docs/`. It parses a composite action or a reusable workflow, pulls out every input and output, and renders them as a markdown table. It's the draft script from this issue, finished off: it now walks the whole repo rather than taking one file at a time, writes the tables into the docs itself, and has tests.

It has four subcommands:

```sh
cd scripts/generate-input-output-docs

go run . generate -root-dir ../../   # rewrite the tables in every doc in the repo
go run . check    -root-dir ../../   # report which docs are stale, without writing
go run . parity   -root-dir ../../   # check a workflow still matches the action it wraps
go run . print -file ../../actions/aws-auth/action.yaml   # dump one file's tables to stdout
```

That PR deliberately changes no action READMEs. Every action here is its own release-please package, so a PR that touches 37 of them would bump 37 versions. Keeping the tool on its own means it lands under `scripts/`, which isn't a package, and triggers no releases.

**The second (#2276) implements the CI check and updates all the docs.** It regenerates every action README and reusable workflow doc from its YAML, and adds `.github/workflows/check-action-docs.yaml`, which regenerates in CI and fails if the committed docs don't match. It's stacked on #1454, so that one merges first.

Between them they cover all three checks from the issue:

---

### 1. `docker-build-push-multiarch.yml` vs `docker-build-push-image/action.yaml`

Done, as `go run . parity`. The rules live in `scripts/generate-input-output-docs/parity.yaml` and run in CI.

[...]

PR #2276, opened 2026-08-27
docs: generate action and workflow input/output tables

### LINKED ISSUE #137: Generate action READMEs from the YAML

ricky-undeadcoders:
Regenerates the Inputs/Outputs tables in every action README and reusable workflow doc from the YAML that declares them, and adds a CI job that fails when they drift.

#1454 added the generator and is now merged; this PR is its output plus the enforcement job. Rebased onto main, so it stands alone.

Titled `docs:` deliberately: this touches 37 action READMEs, and every action is its own release-please package. A `feat:` here would minor-bump all 38 for a documentation change.

## What this catches

The tables now carry `Required` and `Default` generated from the YAML rather than from prose. That immediately surfaced docs that had drifted from their actions:

| Action | Was documented as | Actually |
| --- | --- | --- |
| `cleanup-branches` | `dry-run` defaults to `true` | defaults to `false` — the docs implied a no-op by default, for an action that deletes branches |
| `validate-policy-bot-config` | `policy-bot.grafana.net/api/v1/validate` | `github-policy-bot.grafana-ops.net/api/validate` |
| `reusable-zizmor` | `min-severity` is `medium` | `low` |
| `docker-build-push-multiarch` | output `runner_arches` | `runner-arches` |
| `go-flaky-tests` | 3 outputs | declares none of them |
| `trigger-argo-workflow` | `workflow_template` "defaults to `grafana-ci`" | no such default; the input is required |

21 of 38 actions had drifted. Around 20 previously undocumented inputs/outputs are now covered, and `check-drone-signature` gained a doc it never had.

## How it works

Generated tables live between `<!-- BEGIN_INPUTS -->` / `<!-- END_INPUTS -->` markers. Only the text between markers is replaced, so hand-written prose, footnotes and markdownlint pragmas survive regeneration.

To update after changing an action's YAML:

```sh
cd scripts/generate-input-output-docs
go run . generate -root-dir ../../
```

CI runs the same thing and fails on any diff.

## Source YAML changes

A handful of descriptions were corrected or enriched, because the YAML is now the source of truth and the old hand-written tables carried detail it lacked:

- `docker-import-digests-push-manifest`: `image-digests` was described as a CSV list. It is emitted one per line — the `jq` at `action.yaml:231` produces `"\(.tag)@\(.indexDigest)"` per entry as a multiline output. Corrected, and the gating on `push`/`generate-summary` documented.
- `trigger-argo-workflow`: removed the false "defaults to `grafana-ci`" claim, and fixed an unbalanced quote in `extra_args`.
- `socket-export-sbom`: recorded the `full-scans:create` and `report:read` scopes the token needs.
- `run-capslock`, `generate-openapi-clients`, `find-pr-for-commit`, `push-to-gcs`, `wait-for-docker-publish`, `get-latest-workflow-artifact`, `aws-auth`, `dependabot-auto-triage`, `reusable-zizmor`: restored enum values, conditional-required nuance and other detail that existed only in the old README prose.

## Caught during the rebase

While this sat in review, `create-github-app-token` was rewritten on main
(#2028, node24 + post-job revocation). That changed its inputs, so its generated
table went stale and the drift check flagged it — regenerated here.

Worth noting the rewrite also set `github_app` to `required: true`, which fixes
one of the inaccuracies this work turned up.

## Notes for review

- The drift check uses `git add -A && git diff --staged --exit-code`. Plain `git diff` would miss a doc the generator had to create from scratch, since new files are untracked — and a brand-new action with no README is exactly the case worth catching. (`check-catalog-info.yaml` has this bug today; separate issue to follow.)
- Runs on `ubuntu-x64-small`, matching `sign-and-attest.yml`.
- Bare `<placeholder>` text in descriptions is escaped, otherwise the renderer eats it as an HTML tag and `<image>:<tag>@<digest>` displays as `:@`.

Part of #1466. Closes #137.


actions/get-vault-secrets/README.md
```diff
@@ -50,3 +50,26 @@ jobs:
...
 
 <!-- x-release-please-end-version -->
+
+## Inputs
+
+<!-- BEGIN_INPUTS -->
+
+| Name             | Type    | Required | Default | Description                                                                                                                                                                                                                                                            |
+| ---------------- | ------- | -------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `common_secrets` | String  | No       |         | Common secrets mapping. Only Grafana Labs employees can list these secrets in Vault. These secrets are stored in the `ci/data/common/<secret path>` path in Vault. Ex: `repo_secrets: \| ENVVAR1=secretpath:key ENVVAR2=secretpath:key2`                               |
```
zerok, line 60: Looks like the renderer is struggling with new-lines there. It would be awesome if you/someone could fix this in a follow-up but it's not critical imo.

zerok (dismissed):

dblinkhorn (approved):

actions/get-vault-secrets/README.md
ricky-undeadcoders: Tracking that here: https://github.com/grafana/shared-workflows/issues/2282.

PR #2280, opened 2026-09-08
chore(deps): update zizmor to v1.29.0

kolesnikovae:
Upgrade zizmor to v1.29.0 so required-workflow scans understand GitHub self-repository action references (`uses: $/path/to/action`).

GitHub recommends this syntax for same-repository actions because it resolves from the workflow run commit without depending on the checked-out workspace. Zizmor 1.28.0 rejects the syntax; 1.29.0 adds support for auditing it.

**Testing:** Ran the zizmor collection helper unit tests, parsed the reusable workflow YAML, and audited the workflow with zizmor 1.29.0.


kolesnikovae:
I see that renovate is rate limited and there's a pending zizmor 1.30 update: https://github.com/grafana/shared-workflows/issues/1135. Closing.

PR #2303, opened 2026-09-15
feat(get-vault-secrets): supporting v2 schema

eloymg:
Supporting v2 schema to accessing vault secrets


actions/get-vault-secrets/action.yaml
```diff
@@ -1,6 +1,13 @@
 name: Get Vault Secrets
 description: Composite action (step) to get secrets from Grafana Labs' Vault instance.
 inputs:
+  version:
+    description: |
+      Version of the vault-secrets schema to use. Valid values are `v1` or `v2`.
+      `v1` (default) reads from the shared `ci` mount using the
+      `vault-github-actions` role. `v2` reads from a per-org `ci-<owner>` mount
+      using the `ci-<owner>` role.
```
Proximyst, line 9: This feels like something we should be able to switch over pretty seamlessly, no? Or am I missing something, is `grafana` supposed to migrate from `ci/` too?

eloymg: It's not necessary to migrate, but this way open the possibility to do it in future

Proximyst (approved):

